PR #6, opened 2020-01-29
[WIP] Asynchronous cell execution

davidbrochart:
This PR allows for asynchronous cell execution. I just duplicated (with an `async_` prefix) and made asynchronous the `run_cell` method and all its callers for now.
An option would be to keep only the `async` version of the methods and provide a user-facing (non-async) `execute` method that would accept a `asynchronous=False` parameter. If `False`, `execute` would under the hood emulate a blocking call by running `async_execute` in the event loop with `run_until_complete()`.

### DIFF
--- a/nbclient/execute.py
+++ b/nbclient/execute.py
@@ -3,6 +3,7 @@ from textwrap import dedent
 from contextlib import contextmanager
 from time import monotonic
 from queue import Empty
+import asyncio
 
 from traitlets.config.configurable import LoggingConfigurable
 from traitlets import List, Unicode, Bool, Enum, Any, Type, Dict, Integer, default
@@ -366,6 +367,28 @@ class Executor(LoggingConfigurable):
 
         return self.nb
 
+    # TODO: Remove non-kwarg arguments
+    async def async_execute(self, **kwargs):
+        """
+        Executes each code cell.
+
+        Returns
+        -------
+        nb : NotebookNode
+            The executed notebook.
+        """
+        self.reset_execution_trackers()
+
+        with self.setup_kernel(**kwargs):
+            self.log.info("Executing notebook with kernel: %s" % self.kernel_name)
+            for index, cell in enumerate(self.nb.cells):
+                await self.async_execute_cell(cell, index)
+            info_msg = self._wait_for_reply(self.kc.kernel_info())
+            self.nb.metadata['language_info'] = info_msg['content']['language_info']
+            self.set_widgets_metadata()
+
+        return self.nb
+
     def set_widgets_metadata(self):
         if self.widget_state:
             self.nb.metadata.widgets = {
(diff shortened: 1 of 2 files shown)


nbclient/execute.py
```diff
@@ -575,6 +622,81 @@ def run_cell(self, cell, cell_index=0, store_history=False):
...
+        # aforementioned dropped data.
+
+        # These two variables are used to track what still needs polling:
+        # more_output=true => continue to poll the iopub_channel
+        more_output = True
+        # polling_exec_reply=true => continue to poll the shell_channel
+        polling_exec_reply = True
+
+        while more_output or polling_exec_reply:
+            await asyncio.sleep(0)
```
akhmerov: Out of curiosity, what is the purpose of this await?

davidbrochart: It just yields, so that the event loop can run other tasks.

akhmerov:
I don't quite understand how this works: the part where it makes sense to await is the kernel client events, but we're not actually awaiting on anything there, as far as I can see.

davidbrochart:
The purpose of this PR is just to be able to execute several notebooks in parallel in an async context. Previously it could only be done using threads, because this [while loop](https://github.com/jupyter/nbclient/blob/master/nbclient/execute.py#L529-L573) is blocking.

nbclient/execute.py
```diff
@@ -575,6 +622,81 @@ def run_cell(self, cell, cell_index=0, store_history=False):
...
+            if polling_exec_reply:
+                if self._passed_deadline(deadline):
+                    self._handle_timeout(exec_timeout, cell)
+                    polling_exec_reply = False
+                    continue
+
+                # Avoid exceeding the execution timeout (deadline), but stop
+                # after at most 1s so we can poll output from iopub_channel.
+                timeout = self._timeout_with_deadline(1, deadline)
```
davidbrochart: @akhmerov I think I see what you mean. In practice it allows for some degree of concurrency, because of this 1s timeout, so other cells can be run every second.

davidbrochart:
The user-facing `execute` method calls `async_execute` until it completes (same behavior as before). `async_execute` is an asynchronous method that allows executing several notebooks concurrently in an async context. At the lowest level, we poll for messages and block for 0.1 second so that other cells can be run.
I guess that if we truly want async all the way we should have an async version of `jupyter_client`'s `get_msg` that would return a `Future` from `zmq.asyncio.Socket.poll()`.

davidbrochart:
Trigger CI.

davidbrochart:
@akhmerov although the tests sometimes fail in the CI (due to very slow machines I guess), I think this is ready to be reviewed. Because `jupyter_client`'s `get_msg` uses `zmq.Socket.poll()` which is blocking, I had to poll on `get_msg` in a non-blocking way (no timeout). But once we get `get_msg` to use `zmq.asyncio.Socket.poll()` in `jupyter_client`, we will be able to `await get_msg`. Nevertheless I think this PR demonstrates that by having async execution methods we can already have concurrency without using threads.

davidbrochart:
Closing as #10 has been merged.

PR #9, opened 2020-01-30
DOC: Remove two mentions of Python 2

mgeier:


MSeal:
Test is a random failure with python 3.5 and ipython I've seen for a while :/ Not sure if it affects other python versions.

PR #10, opened 2020-02-02
Async jupyter client

davidbrochart:
This is a follow-up of https://github.com/jupyter/nbclient/pull/6. It has to be used with https://github.com/jupyter/jupyter_client/pull/506.

### DIFF
--- a/nbclient/execute.py
+++ b/nbclient/execute.py
@@ -3,6 +3,7 @@ from textwrap import dedent
 from contextlib import contextmanager
 from time import monotonic
 from queue import Empty
+import asyncio
 
 from traitlets.config.configurable import LoggingConfigurable
 from traitlets import List, Unicode, Bool, Enum, Any, Type, Dict, Integer, default
@@ -366,6 +367,28 @@ class Executor(LoggingConfigurable):
 
         return self.nb
 
+    # TODO: Remove non-kwarg arguments
+    async def async_execute(self, **kwargs):
+        """
+        Executes each code cell.
+
+        Returns
+        -------
+        nb : NotebookNode
+            The executed notebook.
+        """
+        self.reset_execution_trackers()
+
+        with self.setup_kernel(**kwargs):
+            self.log.info("Executing notebook with kernel: %s" % self.kernel_name)
+            for index, cell in enumerate(self.nb.cells):
+                await self.async_execute_cell(cell, index)
+            info_msg = self._wait_for_reply(self.kc.kernel_info())
+            self.nb.metadata['language_info'] = info_msg['content']['language_info']
+            self.set_widgets_metadata()
+
+        return self.nb
+
     def set_widgets_metadata(self):
         if self.widget_state:
             self.nb.metadata.widgets = {
@@ -396,6 +419,30 @@ class Executor(LoggingConfigurable):
             return cell
 
         reply, outputs = self.run_cell(cell, cell_index, store_history)
+        # Backwards compatibility for processes that wrap run_cell
+        cell.outputs = outputs
+
+        cell_allows_errors = self.allow_errors or "raises-exception" in cell.metadata.get(
+            "tags", []
+        )
+
+        if self.force_raise_errors or not cell_allows_errors:
+            if (reply is not None) and reply['content']['status'] == 'error':
+                raise CellExecutionError.from_cell_and_msg(cell, reply['content'])
+
+        self.nb['cells'][cell_index] = cell
+        return cell
+
+    async def async_execute_cell(self, cell, cell_index, store_history=True):
+        """
+        Executes a single code cell.
+
+        To execute all cells see :meth:`execute`.
+        """
+        if cell.cell_type != 'code' or not cell.source.strip():
+            return cell
+
+        reply, outputs = await self.async_run_cell(cell, cell_index, store_history)
         # Backwards compatibility for processes that wrap run_cell
         cell.outputs = outputs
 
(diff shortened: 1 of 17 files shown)


nbclient/execute.py
```diff
@@ -502,14 +532,41 @@ def _passed_deadline(self, deadline):
         return False
 
     def run_cell(self, cell, cell_index=0, store_history=False):
+        loop = get_loop()
+        return loop.run_until_complete(self.async_run_cell(cell, cell_index, store_history))
+
+    async def poll_output_msg(self, parent_msg_id, cell, cell_index, timeout=None):
```
MSeal: Will need to do some more manual testing to ensure we correctly timeout, pass execution context, and handle errors cleanly. Some of those patterns are tested in the synchronous path with some assumptions that might not hold for the async version.

nbclient/tests/test_execute.py
```diff
@@ -270,6 +295,55 @@ def test_many_parallel_notebooks(capfd):
     assert captured.err == ""
 
 
+def test_async_parallel_notebooks(capfd, tmpdir):
```
MSeal: Nice! Glad these tests are here to validate parallel execution

tox.ini
```diff
@@ -33,20 +33,20 @@ commands =
...
-    dist: python3.6
-    docs: python3.6
+    flake8: python3.8
+    manifest: python3.8
+    dist: python3.8
+    docs: python3.8
 deps = .[dev]
-commands = pytest -vv --maxfail=2 --cov=nbclient -W always {posargs}
+commands =
+    /bin/bash -c 'python -m pip install -e git+https://git@github.com/davidbrochart/jupyter_client.git@async_client#egg=jupyter_client'
```
MSeal: We can put that in the requirements file to test as a git path without needing to toss it in the commands. That also makes the package install cleaner for local development and indicates the branch is dependent on an unreleased code dependency.

davidbrochart: I tried to put it in the requirements file but never managed to make it work, that's why it ended up here (I had an error about pip not being able to parse it).

nbclient/execute.py
davidbrochart: Just to explain a bit more how I implemented the asynchronous functionality, there are now two tasks that are run in parallel:
- `_poll_for_reply`: asynchronously awaits a reply on the shell channel (possibly with timeout).
- `poll_output_msg`: asynchronously awaits a message on the IOPub channel for cell execution. We first await as long as no reply is received on the shell channel, and when a reply is received on the shell channel we cancel this task and launch it again with a timeout.

davidbrochart:
Thanks for reviewing @MSeal. There are a couple of issues related to compatibility of async with older Python versions:
- in `setup_kernel` (which is now async because it awaits `start_new_kernel_client` which awaits `wait_for_ready` which awaits `get_msg`...) there is a `yield` which leads to a `SyntaxError: 'yield' inside async function` in Python 3.5. Also in Python 3.5 and 3.6  `asynccontextmanager` doesn't exist in `contextlib`.
- in the tests we now mock messages asynchronously, but `AsyncMock` only exists in `unittest.mock`  from Python 3.8.

That being said, I think we should work on fixing these issues and not have a `nbclient-async` separate module.

SylvainCorlay:
> Another option is to release two modules, nbclient and nbclient-async independently.

That seems a bit overkill to me!

davidbrochart:
Trigger CI.

davidbrochart:
@MSeal I managed to get Python compatibility from 3.5 to 3.8, but it looks like the CI dies randomly for `test_many_parallel_notebooks`. I'm not sure if it's because of low CI resources or if it's a real issue.

davidbrochart:
Trigger CI.

tox.ini
MSeal: all good, there's a way to make it happy but we can instead just focus on getting async support merged and released in jupyter_client as a pre-req for this PR to be released.

nbclient/execute.py
MSeal: Thanks, that helps!

MSeal:
> @MSeal I managed to get Python compatibility from 3.5 to 3.8, but it looks like the CI dies randomly for `test_many_parallel_notebooks`. I'm not sure if it's because of low CI resources or if it's a real issue.

I've seen python 3.5 fail randomly in nbconvert and now in nbclient where the kernel dies unexpectantly due to port conflicts. I believe that's related to ipython no longer releasing 3.5 versions (`ipython==7.9.0` vs `ipython==7.12.0`) and there being some outstanding bugs in ipython 7.9. I think the parallel notebook tests should probably scope to 3.6+ given the ipython issues. However the
```
E       AssertionError: assert '[IPKernelApp] WARNING | Parent appears to have exited, shutting down.\n' == ''
E         - [IPKernelApp] WARNING | Parent appears to have exited, shutting down.
```
is new to me -- that looks like something different is happening.

nbclient/execute.py
```diff
@@ -520,57 +582,14 @@ def run_cell(self, cell, cell_index=0, store_history=False):
...
-                self.process_message(msg, cell, cell_index)
-            except CellExecutionComplete:
-                more_output = False
+        exec_timeout = self._get_timeout(cell)
+        task_poll_output_msg = asyncio.ensure_future(
+            self.poll_output_msg(parent_msg_id, cell, cell_index)
+        )
+        exec_reply = await self._poll_for_reply(parent_msg_id, cell, exec_timeout)
+        if not task_poll_output_msg.done():
+            task_poll_output_msg.cancel()
```
maartenbreddels: Never knew about cancel, can this mean a message can get lost?

davidbrochart: I don't think so, because messages are queued anyway in zmq. And we create a new task polling for messages just after that.

maartenbreddels: But can it cancel, when it's off the zmq queue, say, it's in the json parser. How does it cancel like Ctrl-c.

davidbrochart: It can't do anything between two `await`, so it depends on the implementation. For instance, in `zmq.asyncio.Socket.recv_multipart` we might await `zmq.asyncio.Socket.recv` several times and so cancelling in the middle of it can lead to losing a message, you're right.

SylvainCorlay:
> due to port conflicts

@MSeal there is an accurate description of that issue in https://github.com/jupyter/jupyter_client/issues/487 by @JohanMabille.

Basically the whole handshake model of the kernel protocol is an antipattern. The client should not tell the kernel which ports to use. It should only open one socket for the kernel to communicate back the available ports for itself...

nbclient/execute.py
davidbrochart: Another way that I tried out was to leave the first `poll_output_msg` task running (with no timeout) and launch another one with a timeout, and then `asyncio.wait(both_tasks, return_when=asyncio.FIRST_COMPLETED)`, but it doesn't handle exceptions well (there is a `return_when=asyncio.FIRST_EXCEPTION` for that).
I can't think of a better solution right now.

davidbrochart:
@maartenbreddels I reworked the polling on the shell and iopub channels. Now the polling on iopub channel task is run only once (with no timeout), and when we have an exec reply we `wait_for` that task with a timeout, which will cancel the task after the timeout. That way there is no task rescheduling and no risk to loose a message.

maartenbreddels:
Sounds brilliant, I'll review next week!

(from mobile phone)

davidbrochart:
@MSeal jupyter_client now has an async API (https://github.com/jupyter/jupyter_client/pull/506 has been merged). How would you like to proceed? I guess we should make a release of jupyter_client before merging this PR.

meeseeksmachine:

https://discourse.jupyter.org/t/welcome-new-jupyter-client-nbclient-maintainer-david-brochart/3467/1

PR #15, opened 2020-02-05
Rename executenb() -> execute()

mgeier:
To be perfectly blunt, `executenb()` is a bad name.

### DIFF
--- a/nbclient/tests/test_execute.py
+++ b/nbclient/tests/test_execute.py
@@ -465,7 +465,7 @@ while True: continue
         assert_notebooks_equal(original, executed)
 
     def test_execute_function(self):
-        # Test the executb() convenience API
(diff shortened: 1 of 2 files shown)


nbclient/tests/test_execute.py
```diff
@@ -465,14 +465,14 @@ def process_message(self, msg, cell, cell_index):
         assert_notebooks_equal(original, executed)
 
     def test_execute_function(self):
-        # Test the executenb() convenience API
+        # Test the executb() convenience API
```
MSeal: `execute`

mgeier: Oooops. Fixed in ad4f1880a2d5c929b73f3cc294e252c734e22128.

PR #16, opened 2020-02-05
Remove "execute" sub-module

mgeier:
This changes ...

    nbclient.execute.execute(...)

... into ...

    nbclient.execute(...)

Since it seems that "executing" is the main thing that `nbclient` tends to do, I don't think it's necessary to hide this functionality in a sub-module.

It's quite likely that I missed some places where this would have to be changed as well.


choldgraf:
This seems pretty reasonable to me - I am not super familiar with patterns using `_` before the sub-module in this way, but it I think it's a nifty way to avoid the name collision. Maybe somebody else can comment if this is some kind of anti-pattern I haven't heard of before?

MSeal:
I would not like to have the `_execute.py` file in the repo, it's usually a pattern of "this is private" and I believe would be confusing that we hide all the implementation in a private pattern.

Given the naming concerns, maybe instead we could rename the file to `client.py` and the class to `NotebookClient`? Then the imports would make sense and the class name would match the reponame better?

choldgraf:
I am happy w/ whatever pattern folks are happy with if it allows for `from nbclient import execute` :-)

MSeal:
Thinking over it a few days I think changing the file and class name to reflect the packet would make things more consistent. So you want me to push that change to this PR or make a new one for that @mgeier ?

MSeal:
https://github.com/jupyter/nbclient/pull/17 resolves this issue, going to close for now

mgeier:
@choldgraf 

> I am not super familiar with patterns using `_` before the sub-module in this way, but it I think it's a nifty way to avoid the name collision. Maybe somebody else can comment if this is some kind of anti-pattern I haven't heard of before?

Having `_`-prefixed submodules is a somewhat common pattern, often the files are called `_core.py` or `_base.py` or something along those lines and their "public" parts are imported in `__init__.py`.

A random example: https://github.com/python/cpython/blob/master/Lib/concurrent/futures/_base.py

An alternative pattern is to simply write the whole implementation into `__init__.py`.

A random example for that: https://github.com/python/cpython/blob/master/Lib/venv/__init__.py

Neither of them is an anti-pattern.

Yet another option would be to avoid "packages" (i.e. directories) and sub-modules altogether and implement everything in a single top-level module. This of course only makes sense for reasonably small projects.

@MSeal 

> I would not like to have the `_execute.py` file in the repo, it's usually a pattern of "this is private" and I believe would be confusing that we hide all the implementation in a private pattern.

Yes, the sub-module is indeed private (nobody should use it directly), but you can selectively import the parts you want to be public into the parent module.

PR #17, opened 2020-02-10
Renamed Executor to NotebookClient for better clarity with package

MSeal:
Based on conversation in https://github.com/jupyter/nbclient/pull/16 -- it seems a good idea to rename the file / class we've implemented before the first release.


MSeal:
@mgeier This also allows for the import pattern you were asking for `from nbclient import execute`.

MSeal:
Yep just cosmetic :)

Issue #14, opened 2020-02-05
Why isn't this project called "nbexecute"?


mgeier:
I was wondering what `nbclient` is supposed to do except executing ... and I asked there: https://github.com/jupyter/nbclient/issues/4#issuecomment-579499964

This was answered by @MSeal in https://github.com/jupyter/nbclient/issues/4#issuecomment-581790592:

> For now the intention is to have a very narrow scoped notebook executor that has few dependencies, like jupyter_client does for interacting with kernels. This library doesn't handle output IO and does one thing well. It'll likely get an async execution pattern, but beyond that I don't see major responsibilities being added to the core the library for now.

So if it is only ever supposed to "execute", why call it by a totally un-intuitive and confusing name?

choldgraf:
I think it's reasonable to bring up the name of the project as a topic for discussion. I agree that `nbexecute` would be more intuitive for me, but I am not sure if there were other reasons that `nbclient` was chosen as the name. Either way it is not something I feel super strongly about

---

That said, please use a more constructive tone when you're engaging with this community - calling something confusing and un-intuitive is an opinion, not a factual statement, and blanket statements like this carry with it a negative connotation that may not lead to constructive discussion. For example, another way to say the same thing is "The name `nbclient` feels unintuitive to me, and may be to others". (I won't mention it there, but the same thing applies to #15). Please keep comments neutral-to-positive, and constructive.

MSeal:
My original idea was to name this nbexecute. I shopped the naming around with a number of the jupyter core devs and nbclient was the name most voted for. The parallel to jupyter_client for responsibilities was what resonated for folks.

Here's the list of names I ended up sending to folks:

- nbexec
- nbexecute
- nbproc  
- nbclient (plays ontop of the jupyter_client naming theme)
- nbscribe
- scribbles (bottom of the barrel)

I'm in a similar line of thought as @choldgraf , but given the naming was already picked among a few core devs I'd prefer to not change it unless it was deemed as problematic or many people thought it was a bad fit.

But you know... "There are only two hard things in Computer Science: cache invalidation and naming things" - Phil Karlton

choldgraf:
One thought here - we could include a short description of the reasoning behind the name in the docs. At the least it could help users remember the name...is there a public place where the `nbclient` naming conversation happened?

MSeal:
[...]

PR #21, opened 2020-02-14
Hide implementation details

mgeier:
Since the underlying "client" infrastructure seems to be less than ideal (see https://github.com/jupyter/nbclient/issues/14#issuecomment-584514077), it might be better to keep the public API to a minimum for now.

This way, the implementation can freely evolve and once a nice API is reached, it can be made public without breakage.

Probably some of the API in the `exceptions` sub-module should also made private for now?

[UPDATE: currently the docs build is broken because I didn't update them. I'm waiting for some feedback and then I can update the docs]


MSeal:
I don't really agree with hiding the client -- it's the object downstream libraries can extend for specialized effect (e.g. papermill 2.0 and soon nbconvert 6.0)

MSeal:
Closing this as we're exposing and leaning on the interface that's emerging. Having the class that's extensible has been very useful for downstream projects so far.

Issue #20, opened 2020-02-11
Make task for copying changelogs to readthedocs page


MSeal:
Either create an automated task to copy and convert the markdown changelogs to the readthedocs /docs path or just move the changelog to rst and place it there (I dislike writing rst by hand).

choldgraf:
Why not use recommonmark to parse the changelog in markdown? Usually I use `CHANGELOG.md` in the root that just has a link to the changelog page in the documentation.

MSeal:
That sounds like a good path to try here. If no one gets to it first I'll give it a try before the next release

PR #22, opened 2020-02-15
adding changelog in docs

choldgraf:
closes https://github.com/jupyter/nbclient/issues/20

@MSeal does this look reasonable?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,3 @@
 # Change Log
 
-## 0.1.0
-- Initial release -- moved out of nbconvert 6.0.0-a0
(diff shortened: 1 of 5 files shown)


MSeal:
Docs builds failed with an odd error. Looks almost like it just failed talking to PyPI

MSeal:
I restarted the jobs

choldgraf:
I think the errors were because we had pinned to an older version of recommonmark...but I just removed the pin and it seems to be happy. Is there a strong reason we were using the 0.4 version before?

MSeal:
I believe there was a bug with the most recent version or a python 2 specific reason (this was copied from another project), but I can't recall. If it works now to unpin +1

PR #32, opened 2020-02-28
Add execution timing in code cell metadata for v4 spec

davidbrochart:


davidbrochart:
I added a test which checks for expected timing values and format. I think it's ready to be merged now.

Issue #33, opened 2020-03-08
Async not awaited in some tests. 


MSeal:
Looks like it's mostly benign around test mocking not being updated to await always, but we probably want to remove the warnings before releasing in case there's a bug associated? Thoughts @davidbrochart 

```
=============================== warnings summary ===============================
nbclient/tests/test_client.py::TestRunCell::test_deadline_iopub
  /usr/lib/python3.6/unittest/mock.py:684: RuntimeWarning: coroutine 'make_async.<locals>._' was never awaited
    return object.__setattr__(self, name, value)

nbclient/tests/test_client.py::TestRunCell::test_display_data_message
nbclient/tests/test_client.py::TestRunCell::test_unknown_comm_message
  /usr/lib/python3.6/unittest/mock.py:1898: RuntimeWarning: coroutine 'make_async.<locals>._' was never awaited
    self.name = name

nbclient/tests/test_client.py::TestRunCell::test_eventual_deadline_iopub
  /home/mseal/Workspace/nbclient/nbclient/tests/test_client.py:719: RuntimeWarning: coroutine 'make_async.<locals>._' was never awaited
    message_mock.side_effect = message_seq(list(message_mock.side_effect)[:-1])

nbclient/tests/test_client.py::TestRunCell::test_eventual_deadline_iopub
  /home/mseal/.py3local/lib/python3.6/site-packages/traitlets/traitlets.py:510: RuntimeWarning: coroutine 'make_async.<locals>._' was never awaited
    if meth_name in cls.__dict__:
```

MSeal:
Hmm looks like it might be an actual issue when I was testing with papermill (some changes needed to make that work as well):

```
sys:1: RuntimeWarning: coroutine 'NotebookClient._wait_for_reply' was never awaited
Executing: 100%|████████████████████████████████████████████████████████████████████████████████████████████| 3/3 [01:04<00:00, 10.02s/cell]
Traceback (most recent call last):
  File "/home/mseal/.py3local/bin/papermill", line 11, in <module>
    load_entry_point('papermill', 'console_scripts', 'papermill')()
  File "/home/mseal/.py3local/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/mseal/.py3local/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
[...]

Issue #31, opened 2020-02-28
Release


davidbrochart:
When do you plan to make a release? We need one in order to have async execution in Voila (https://github.com/voila-dashboards/voila/pull/541).
Also, I saw that nbclient was released in PyPI (https://pypi.org/project/nbclient/#history) but not in the GitHub repo.

MSeal:
There's one issue that came up internally here in testing papermill 2.0 and nbclient around nested papermill calls from subprocesses that might be related to nbclient holding onto kernel state too long. I'm going to test for that more this weekend and submit a fix if it is the root cause (haven't made an issue for it yet). Other than that I think we can do a release unless there's anything you know about that we need to address first.

> Also, I saw that nbclient was released in PyPI (https://pypi.org/project/nbclient/#history) but not in the GitHub repo.

Whoops, fixed now.

choldgraf:
I am also not super sure whether or not I've uncovered a bug in #7 - if that's a bug, it should probably be a blocker on a release as well

MSeal:
I can dig into that one as well, I think it's likely a config issue but yes we should figure that out first

MSeal:
I got a little sick this weekend, will dig in and help get a release out after I'm feeling better.

choldgraf:
@MSeal don't sweat it, and I hope you feel better!

MSeal:
Release is out finally! Thanks everyone for the hard work

davidbrochart:
Thanks a lot Matthew!

choldgraf:
hooray!!! 🎉🎉🎉

congrats everybody, and thanks @MSeal for pushing us over the finish line :-)

davidbrochart:
Do we want to tag on GitHub too? I only see 0.1.0 in https://github.com/jupyter/nbclient/releases

MSeal:
Done. I thought I had pushed the tags too but I guess I missed that.

choldgraf:
We all owe you a 🍺 @MSeal

PR #34, opened 2020-03-09
Zombie Process Fix

MSeal:
This PR (probably) fixes both the dead / zombie process issues that come up when calling nbclient in nested subprocesses as well as the memory leak when running nbclient repeatidely in the same process (might fix https://github.com/nteract/papermill/issues/478). These were both being caused by zmq processes not being cleaned up, which the change in moving to nbclient left around by not deleting the kernel objects after execution. The cleanup process has been made much more robust to a number of failure modes and will attempt a direct kill call on the child process if graceful shutdown is not achieved.

Also fixed a number of async issues / warnings that were cropping up. Downstream libraries were relying on some functions being synchronous, so I renamed a number of functions to `async_` and used loop calls to avoid compatibility issues for the next release.

Most of these were really hard to write a meaningful unittest. I made some notebooks that recreated complex setups that caused process issues (or emitted a warning that would do so in some circumstances) and cleaned things up until warnings and hanging processes were no long around. I don't think there's an easy way to encode that into unittests unfortunately.

Needed for moving forward with https://github.com/jupyter/nbclient/issues/31
Fixes runtime issues, but not all test warnings for https://github.com/jupyter/nbclient/issues/33

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -5,6 +5,7 @@ from textwrap import dedent
 # For python 3.5 compatibility we import asynccontextmanager from async_generator instead of
 # contextlib, and we `await yield_()` instead of just `yield`
 from async_generator import asynccontextmanager, async_generator, yield_
+from contextlib import contextmanager
 
 from time import monotonic
 from queue import Empty
@@ -15,8 +16,14 @@ from traitlets import List, Unicode, Bool, Enum, Any, Type, Dict, Integer, defau
 
 from nbformat.v4 import output_from_msg
 
-from .exceptions import CellTimeoutError, DeadKernelError, CellExecutionComplete, CellExecutionError
-from .util import run_sync
+from .exceptions import (
+    CellControlSignal,
+    CellTimeoutError,
+    DeadKernelError,
+    CellExecutionComplete,
+    CellExecutionError
+)
+from .util import run_sync, await_or_block
 
 
 def timestamp():
@@ -324,7 +331,28 @@ class NotebookClient(LoggingConfigurable):
         self.km.client_class = 'jupyter_client.asynchronous.AsyncKernelClient'
         return self.km
 
-    async def start_new_kernel_client(self, **kwargs):
+    async def _async_cleanup_kernel(self):
+        try:
+            # Send a polite shutdown request
+            await await_or_block(self.kc.shutdown)
+            try:
+                # Queue the manager to kill the process, sometimes the built-in and above
+                # shutdowns have not been successful or called yet, so give a direct kill
+                # call here and recover gracefully if it's already dead.
+                await await_or_block(self.km.shutdown_kernel, now=True)
+            except RuntimeError as e:
+                # The error isn't specialized, so we have to check the message
+                if 'No kernel is running!' not in str(e):
+                    raise
+        finally:
+            # Remove any state left over even if we failed to stop the kernel
+            await await_or_block(self.km.cleanup)
+            await await_or_block(self.kc.stop_channels)
+            self.kc = None
+
+    _cleanup_kernel = run_sync(_async_cleanup_kernel)
+
+    async def async_start_new_kernel_client(self, **kwargs):
         """Creates a new kernel client.
 
         Parameters
@@ -346,22 +374,44 @@ class NotebookClient(LoggingConfigurable):
         if self.km.ipykernel and self.ipython_hist_file:
             self.extra_arguments += ['--HistoryManager.hist_file={}'.format(self.ipython_hist_file)]
 
-        await self.km.start_kernel(extra_arguments=self.extra_arguments, **kwargs)
+        await await_or_block(self.km.start_kernel, extra_arguments=self.extra_arguments, **kwargs)
 
         self.kc = self.km.client()
-        self.kc.start_channels()
+        await await_or_block(self.kc.start_channels)
         try:
-            await self.kc.wait_for_ready(timeout=self.startup_timeout)
+            await await_or_block(self.kc.wait_for_ready, timeout=self.startup_timeout)
         except RuntimeError:
-            self.kc.stop_channels()
-            await self.km.shutdown_kernel()
+            await self._async_cleanup_kernel()
             raise
         self.kc.allow_stdin = False
         return self.kc
 
+    start_new_kernel_client = run_sync(async_start_new_kernel_client)
+
+    @contextmanager
+    def setup_kernel(self, **kwargs):
+        """
+        Context manager for setting up the kernel to execute a notebook.
+
+        The assigns the Kernel Manager (`self.km`) if missing and Kernel Client(`self.kc`).
+
+        When control returns from the yield it stops the client's zmq channels, and shuts
+        down the kernel.
+        """
+        # Can't use run_until_complete on an asynccontextmanager function :(
+        if self.km is None:
+            self.start_kernel_manager()
+
+        if not self.km.has_kernel:
+            self.start_new_kernel_client(**kwargs)
+        try:
+            yield
+        finally:
+            self._cleanup_kernel()
+
     @asynccontextmanager
     @async_generator  # needed for python 3.5 compatibility
-    async def setup_kernel(self, **kwargs):
+    async def async_setup_kernel(self, **kwargs):
         """
         Context manager for setting up the kernel to execute a notebook.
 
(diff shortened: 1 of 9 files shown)


nbclient/client.py
```diff
@@ -308,7 +316,49 @@ def start_kernel_manager(self):
...
+            loop = get_loop()
+            try:
+                # Queue the manager to kill the process, sometimes the built-in and above
+                # shutdowns have not been successful or called yet, so give a direct kill
+                # call here and recover gracefully if it's already dead.
+                try:
+                    # For AsyncKernelManager
+                    loop.run_until_complete(self.km.shutdown_kernel(now=True))
+                except TypeError:
+                    self.km.shutdown_kernel(now=True)
```
davidbrochart: Should we check if `self.km.shutdown_kernel` is a coroutine, instead of using `try/except`?
Also, we might want to have an async version of `_cleanup_kernel`, which is use in blocking and async functions.

nbclient/client.py
```diff
@@ -330,22 +380,47 @@ def start_kernel_manager(self):
         if self.km.ipykernel and self.ipython_hist_file:
             self.extra_arguments += ['--HistoryManager.hist_file={}'.format(self.ipython_hist_file)]
 
-        self.km.start_kernel(extra_arguments=self.extra_arguments, **kwargs)
+        # Support AsyncKernelManager
+        if inspect.iscoroutinefunction(self.km.start_kernel):
```
davidbrochart: We could also use `asyncio.iscoroutinefunction`, but I guess this is fine.

nbclient/client.py
```diff
@@ -330,22 +380,47 @@ def start_kernel_manager(self):
...
+            self.km.start_kernel(extra_arguments=self.extra_arguments, **kwargs)
 
         self.kc = self.km.client()
         self.kc.start_channels()
         try:
             await self.kc.wait_for_ready(timeout=self.startup_timeout)
         except RuntimeError:
-            self.kc.stop_channels()
-            self.km.shutdown_kernel()
+            self._cleanup_kernel()
```
davidbrochart: Here we might use an async version of `_cleanup_kernel`, since we are in an async function.

nbclient/client.py
```diff
@@ -358,12 +433,11 @@ def start_kernel_manager(self):
...
 
         if not self.km.has_kernel:
-            await self.start_new_kernel_client(**kwargs)
+            await self.async_start_new_kernel_client(**kwargs)
         try:
             await yield_(None)  # would just yield in python >3.5
         finally:
-            self.kc.stop_channels()
-            self.kc = None
+            self._cleanup_kernel()
```
davidbrochart: Here we might use an async version of `_cleanup_kernel`, since we are in an async function.

nbclient/exceptions.py
```diff
@@ -1,4 +1,13 @@
-class CellTimeoutError(TimeoutError):
+class CellControlSignal(Exception):
+    """
+    A custom exception used to indicate that the exception is used for cell
+    control actions (not the best model, but it's needed to cover existing
+    behvior without major refactors).
```
davidbrochart: Typo `behvior` -> `behavior`.

nbclient/client.py
```diff
@@ -507,19 +584,25 @@ def _check_alive(self):
...
 
-    async def _wait_for_reply(self, msg_id, cell=None):
+    def _wait_for_reply(self, msg_id, cell=None):
+        # Needed by papermill until it updates to consume async_wait
+        loop = get_loop()
+        return loop.run_until_complete(
+            self.async_wait_for_reply(msg_id, cell=cell)
+        )
+
+    async def async_wait_for_reply(self, msg_id, cell=None):
```
davidbrochart: Should it be `_async_wait_for_reply` for name consistency with `_wait_for_reply`?

davidbrochart:
Thanks for this PR @MSeal.
I am wondering if we should have a `blocking` and an `asynchronous` directory, as it is for `jupyter_client`, to clearly separate the two implementations?

nbclient/client.py
MSeal: Based on https://stackoverflow.com/questions/36076619/test-if-function-or-method-is-normal-or-asynchronous the `asyncio` is probably a better choice to cover more cases. I'll change em over.

nbclient/client.py
MSeal: :+1: will do

nbclient/client.py
MSeal: I was thinking of exposing `async_wait_for_reply` as not private so it could be used in contracts for extensibility since papermill ended up needing the blocking verison to extend behavior. If I'm doing that I should probably commit and make `wait_for_reply`, or revert the decision overall and keep with `_async_wait_for_reply`.

MSeal:
> Thanks for this PR @MSeal.
I am wondering if we should have a blocking and an asynchronous directory, as it is for jupyter_client, to clearly separate the two implementations?

We could do that (or rather have two files, asyncclient.py and client.py would be simple enough given the lack of directory structure here). Only downside is that we may end up repeating some logic that we're currently wrapping in run_until_complete. It would be a be cleaner and we could have two distinct classes that clearly separate async from blocking concerns... Might need a base class to hold common synchronous code patterns. What's you preference @davidbrochart ?

Also might be a couple days before responding on this PR fyi

davidbrochart:
I've made some refactoring along this line in #37.

MSeal:
Assuming #37 merges, do you want me to address separating the class into blocking / non-blocking versions or finish providing lower level interfaces in each to blocking and non-blocking functions to be used by downstream extensions still?

nbclient/client.py
MSeal: I can change to that pattern if you prefer.

davidbrochart:
I think what we now have in #37 with blocking methods using `run_sync` on async methods is fine.

davidbrochart:
I believe this is blocking the release. Do you need help @MSeal ?

MSeal:
I've got time to take a pass tonight on updating. Should be able to more quickly respond / finish up the release now that some other threads are resolved.

MSeal:
See how those changes look. Had to do more than I expected unfortunately so the diff is longer

nbclient/util.py
```diff
@@ -45,3 +47,17 @@ def wrapped(self, *args, **kwargs):
...
+    on execution.
+    """
+    if asyncio.iscoroutinefunction(func):
+        return await func(*args, **kwargs)
+    else:
+        result = func(*args, **kwargs)
+        # Mocks mask that the function is a coroutine :/
+        if isinstance(result, Coroutine):
+            return await result
+        return result
```
davidbrochart: In jupyter_server, we will have something similar that we call [ensure_async](https://github.com/jupyter/jupyter_server/blob/a2610d33be535b42e37e3c200bcec6afeec341d8/jupyter_server/utils.py#L386-L400). We use it as a wrapper on a function call when we don't know it this function is async or not. I prefer that over passing the function object and its parameters to `await_or_block` and let it make the call. I will try it and maybe make a PR on your `zombieProcFix` branch.

MSeal: I'm ok with that change, thanks for helping

MSeal:
I haven't gotten a chance to dig into the failures -- Looks like something is trying to await a non-async call for at least one of them.

davidbrochart:
I sent a PR on your branch: https://github.com/MSeal/nbclient/pull/2

MSeal:
:doh: -- I added that fork as a watched repo so I see those changes in a timely manner going forward.

MSeal:
I think we may need to drop 3.5 support if async keeps hanging with it :/. Same failures are occurring in the nbconvert port to use nbclient with python 3.5. End of life is around the corner 2020-09-13, maybe we should just pull the trigger now and drop it.

@captainsafia @willingc I checked some known images / VMs that were pinned to 3.5 and those I know about have moved the defaults forward to 3.6 or 3.7. Is there internals you know of that depend heavily on 3.5? If we drop support here / in jupyter_client (where I think the issue also manifesting) it would mean papermill, nbconvert, and scrapbook would have to follow to 3.6+.

captainsafia:
> Is there internals you know of that depend heavily on 3.5? If we drop support here / in jupyter_client (where I think the issue also manifesting) it would mean papermill, nbconvert, and scrapbook would have to follow to 3.6+.

Based on my inquiries, it seems like most people are on 3.6 already.

MSeal:
I asked around a bit more to confirm the change would generally be accepted. I'm going to go forward with dropping 3.5 support so we don't mislead people into failing processes.

MSeal:
@davidbrochart Good with a merge / release?

davidbrochart:
BTW, do you need help in the future for the releases?

MSeal:
@davidbrochart That would be great. I'll get this one as PyPI permissions are more restrictive than github for jupyter project.

I would really like to have more python developers doing release management across the projects so I'm not a bottleneck. I'll get that conversation going about promoting PyPI permissions to more people with the core team.

MSeal:
@davidbrochart Actually if you wanted to help with this one, we could use an announcement to discourse and the jupyter mailing list. You did the lion share of the async work and it'd be great to publicize the changes in the release there. Up to you, I can also write it up

davidbrochart:
@MSeal sure, let me try.

PR #35, opened 2020-03-09
Fix nested event loops execution

davidbrochart:
Fixes `RuntimeError: This event loop is already running` when `nbclient.execute` is run in an environment where an event loop is already running (e.g. Binder). See https://github.com/jupyter/nbclient/pull/7#issuecomment-592109215.


choldgraf:
As we don't have the Binder repo merged yet it's a bit hard to know if that fixes the problem, but I did try running @davidbrochart 's branch on Binder here:

https://mybinder.org/v2/gh/davidbrochart/nbclient/56a1b042499673026759f43807e6169a69541aae

and it worked! So +1 from me

PR #37, opened 2020-03-10
Make nest_asyncio optional

davidbrochart:
In nbclient, we chose to implement blocking methods such as `execute` by running their async counterparts (e.g. `async_execute`) in an event loop (with `asyncio.run_until_complete`), to avoid code duplication. This can be an issue in environments where an event loop is already running, such as a Jupyter Notebook. [nest_asyncio](https://github.com/erdewit/nest_asyncio) solves this issue, but it doesn't play well with tornado (see https://github.com/tornadoweb/tornado/issues/2753). This will impact applications such as Voila (which will use the async methods of nbclient and so won't need to nest event loops). We should for now make the use of nest_asyncio an opt-in.

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -294,6 +294,34 @@ class NotebookClient(LoggingConfigurable):
         self.km = km
         self.reset_execution_trackers()
 
+    def run_blocking(self, coro):
+        """Runs a coroutine and blocks until it has executed.
+
+        An event loop is created if no one already exists. If an event loop is
+        already running, this event loop execution is nested into the already
+        running one if `nest_asyncio` is set to True.
+
+        Parameters
+        ----------
+        coro : coroutine
+            The coroutine to be executed.
+
+        Returns
+        -------
+        result :
+            Whatever the coroutine returns.
+        """
+        try:
+            loop = asyncio.get_event_loop()
+        except RuntimeError:
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
+        if self.nest_asyncio:
+            import nest_asyncio
+            nest_asyncio.apply(loop)
+        result = loop.run_until_complete(coro)
+        return result
+
     def reset_execution_trackers(self):
         """Resets any per-execution trackers.
         """
@@ -379,6 +407,9 @@ class NotebookClient(LoggingConfigurable):
         finally:
             self.kc.stop_channels()
             self.kc = None
+
+    def execute(self, **kwargs):
+        return self.run_blocking(self.async_execute(**kwargs))
 
     async def async_execute(self, **kwargs):
         """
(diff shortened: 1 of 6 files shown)


choldgraf:
Correct me if I'm wrong, but this is also removing one of the more-common user-facing methods `execute`, no?

davidbrochart:
No, but I agree it looks like it. It's actually monkey-patched with the `make_blocking` function. Not sure this is acceptable though.

davidbrochart:
I removed the monkey-patching of `NotebookClient`. Now `execute` and `execute_cell` are explicitly declared and call `run_blocking` on their async counterparts.

davidbrochart:
Trigger CI.

davidbrochart:
Trigger CI.

choldgraf:
Another user-facing question from me - does it mean that running this from within Jupyter Notebooks will require using the `next_asyncio` flag? I feel like this will be not be obvious to people. Is there any way to check whether we need to nest based on environment variables etc? Or a way to catch the error message and display something like "try running with `next_asyncio = True`?

davidbrochart:
Thanks for you comment @choldgraf, this last commit shows a meaningful message when nest event loops seems to be needed.

choldgraf:
Nice - I like that, I think it'll be helpful for people that aren't familiar with some of these issues (which unfortunately is probably exactly the kind of person that would be calling `execute` from within a notebook environment)

davidbrochart:
I'm going to merge this PR soon, if nobody has objections.

nbclient/client.py
```diff
@@ -368,20 +418,11 @@ def start_kernel_manager(self):
             self.kc = None
 
     def execute(self, **kwargs):
-        """
```
MSeal: I think these doc strings are useful, probably should leave them in.

nbclient/client.py
```diff
@@ -551,47 +592,13 @@ def _check_raise_for_error(self, cell, exec_reply):
                 raise CellExecutionError.from_cell_and_msg(cell, exec_reply['content'])
 
     def execute_cell(self, cell, cell_index, execution_count=None, store_history=True):
-        """
```
MSeal: ditto on docstring

davidbrochart:
Welcome back Matthew, I hope you are fine.
The last commit refactors the previous `run_blocking` into `run_sync`, which automatically inherits the docstring of the async method. Also, `is_alive` is now async in jupyter_client.

davidbrochart:
The failing tests are related to https://github.com/jupyter/nbclient/pull/38#issuecomment-598703625, which will be solved when we have a jupyter_client release.
Also `test_many_parallel_notebooks` randomly fails, we need to investigate that.

MSeal:
> Welcome back Matthew, I hope you are fine.

Yep all good so far, went through some exposed places getting home but no symptoms as of 3 days in. Disconnected from my inbox for a bit was all.

davidbrochart:
I will merge this PR soon, if no objection.

MSeal:
merged

PR #38, opened 2020-03-13
Use AsyncKernelManager from jupyter_client

davidbrochart:


davidbrochart:
Not sure why `pip` doesn't want to install jupyter_client from GitHub for python3.6. Next release of jupyter_client will include `AsyncKernelManager` and so we will be able to install from PyPI anyway.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 traitlets>=4.2
-jupyter_client>=6.0.0
```
MSeal: Let's wait for 6.1.0 before we merge so we have proper versions in the dependency files

MSeal: ok should be changeable to 6.1.0 now

PR #44, opened 2020-04-09
Add reset_kc option to reset_execution_trackers

davidbrochart:
`reset_kc` can also be passed to `async_execute` and `execute` methods (defaults to False: the kernel client won't be reset).

### DIFF
--- a/nbclient/tests/test_client.py
+++ b/nbclient/tests/test_client.py
@@ -558,14 +558,14 @@ while True: continue
         executor.execute()
         # we didn't ask to reset the kernel client, a new one must have been created
         kc = executor.kc
-        self.assertNotEqual(kc, None)
+        assert kc is not None
         executor.execute()
         # we didn't ask to reset the kernel client, the previously created one must have been reused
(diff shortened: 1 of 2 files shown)


MSeal:
Mind add a test that the kc gets reused using this flab on subsequent execute calls?

davidbrochart:
Thank you for reminding me the importance of tests, it revealed a bug.

nbclient/tests/test_client.py
```diff
@@ -544,6 +544,29 @@ def test_force_raise_errors(self):
...
+
+        executor = NotebookClient(
+            input_nb,
+            resources=self.build_resources(),
+        )
+
+        executor.execute()
+        # we didn't ask to reset the kernel client, a new one must have been created
+        kc = executor.kc
+        self.assertNotEqual(kc, None)
```
maartenbreddels, line 561: I guess we're using pytest right? then we can use this?
```suggestion
        assert kc is not None
```

davidbrochart: Yes.

PR #46, opened 2020-04-21
fix nbclient description in docs

rohitsanj:


MSeal:
Thank you for fixing that copy paste error

PR #51, opened 2020-05-12
Support MultiKernelManager

davidbrochart:
In order to remove [some boilerplate code in Voila](https://github.com/voila-dashboards/voila/blob/1c9ca89d5915db42179755fc0936b2f9cdd314cf/voila/handler.py#L116-L138), it would be nice if `NotebookClient` could accept a `MultiKernelManager` instead of a `KernelManager`. Their APIs differ, but a `MultiKernelManager`'s underlying `KernelManager` can be retrieved through the `get_kernel(kernel_id)` method.
This PR makes `(async_)start_new_kernel_client` support a `MultiKernelManager` by checking the presence of a `client()` method, which only a `KernelManager` has.


davidbrochart:
Trigger CI.

davidbrochart:
Trigger CI.

PR #53, opened 2020-05-14
Adding reset_kc to setup_kernel

danielfrg:
Matching what async_setup_kernel does.

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -411,7 +411,7 @@ class NotebookClient(LoggingConfigurable):
         When control returns from the yield it stops the client's zmq channels, and shuts
         down the kernel.
         """
-        reset_kc = kwargs.pop('reset_kc', False)
+        cleanup_kc = kwargs.pop('cleanup_kc', True)
 
         # Can't use run_until_complete on an asynccontextmanager function :(
         if self.km is None:
@@ -422,7 +422,7 @@ class NotebookClient(LoggingConfigurable):
         try:
             yield
         finally:
-            if reset_kc:
+            if cleanup_kc:
                 self._cleanup_kernel()
 
     @asynccontextmanager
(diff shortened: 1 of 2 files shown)


nbclient/client.py
```diff
@@ -411,6 +411,8 @@ def setup_kernel(self, **kwargs):
         When control returns from the yield it stops the client's zmq channels, and shuts
         down the kernel.
         """
+        reset_kc = kwargs.pop('reset_kc', False)
```
MSeal, line 414: I'd rename this to `cleanup_kc` and change `def setup_kernel(self, **kwargs):` to `def setup_kernel(self, cleanup_kc=False, **kwargs):`

MSeal: Ahh I see you just copied the async kernel implementation -- we should probably clean that up as well to make it clearer how it works, but I can also see the argument for keeping the current convention to avoid breaking people.

danielfrg:
I will add a test soon, I edited this directly on GitHub 🙈

Let me know what you prefer in terms of name/functionality.
My opinion: I agree that `cleanup_kc` is better.

I can also clean up `async_setup_kernel` as part of this, just let me know.

If you want to maintain the current convention I guess the default should be `cleanup_kc =True` by default, thats what happens now without the flag and seams reasonable to be because it's a contextmanager.

MSeal:
@davidbrochart what do you think about changing `reset_kc` to `cleanup_kc`?

davidbrochart:
Yes it is more coherent to change it to `cleanup_kc`. It's relatively new so I think we can afford that change.

danielfrg:
I cleaned changed `reset_kc` to `cleanup_kc` on the `*setup_kernel` functions and added a `test_cleanup_kernel_client` test.

Cleaned up a bit `reset_kc` on `execute`.

I think this should make it more coherent.

PR #61, opened 2020-05-19
Fixing AttributeError when kernel cleanup occurs before the kernel is created.

golf-player:


golf-player:
I put the check in execute since it's only the first one causing problems. In other cases where kernel cleanup is called and fails because kernel_manager is not defined, it's probably best to fail since something is very wrong

davidbrochart:
LGTM.

davidbrochart:
Thanks!

golf-player:
Appreciate the quick response. Thanks!

PR #63, opened 2020-05-19
Use GH actions to test on OSX and Windows as well

davidbrochart:


davidbrochart:
Some interesting results: https://github.com/davidbrochart/nbclient/pull/1

choldgraf:
Haha whoa - I wonder why it isn't running on this PR though?

davidbrochart:
Yes I don't know why either. And it doesn't run the `coverage` job.

choldgraf:
what if you manually set it to `on: pull_request, push`? Maybe it's the `*` that is the issue?

davidbrochart:
Ah it needs the tests to pass before doing code coverage :smile:

davidbrochart:
`test_many_parallel_notebooks` was failing because it uses multiprocessing (a pool of 2 cores +  the main python program). Maybe we only have one core on GH actions?

davidbrochart:
This is ready to be merged. We are keeping Travis CI anyway, so it's just adding GitHub actions CI, except that we don't use `pebble` anymore for `test_many_parallel_notebooks` (for some reason it didn't work on GH actions, and we have the same with `concurrent.futures.ProcessPoolExecutor`).

PR #65, opened 2020-05-20
Do not lose cause of exception

maartenbreddels:


choldgraf:
I've never seen this syntax before - what does it do?! I am intrigued...

I am hesitant to merge since I don't really know what the ramification is, but somebody else feel free to do so if you're +1

choldgraf:
that is so cool!

PR #67, opened 2020-05-25
[WIP] Run coroutines in another thread for blocking API

davidbrochart:
Trying to get rid of nest_asyncio (see https://github.com/nteract/papermill/issues/490).

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -101,21 +101,6 @@ class NotebookClient(LoggingConfigurable):
         ),
     ).tag(config=True)
 
-    nest_asyncio = Bool(
-        False,
-        help=dedent(
-            """
-            If False (default), then blocking functions such as `execute`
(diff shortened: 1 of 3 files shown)


davidbrochart:
@MSeal this seems like the deadlock issue you were mentioning, `test_many_parallel_notebooks` hangs and it is the test which does multiprocessing.

MSeal:
Yep -- this is basically guaranteed with os threading ontop of forking at the same time because, among other reasons, the os replicates thread local context that's invalid.

MSeal:
Is there anyway we can apply or not apply nest_asyncio and catch the exception to switch modes automatically instead of introducing more concurrent patterns in-place?

davidbrochart:
Yes, we can try without nest_asyncio and if we catch the exception try again with it. Thanks, I'll open a PR.

Issue #24, opened 2020-02-16
Widget Output objects save wrong display content


MSeal:
See https://github.com/jupyter/nbconvert/issues/923 for root-cause analysis and proposal for solution of the issue for the code base when it was in nbconvert.

choldgraf:
I just want to confirm that this is still an unresolved issue, right? I know we've had some changes in nbclient since this opened up but none of that resolve this yeah?

MSeal:
Yep still not solved, too many projects :) It's in my queue of things to go do but my free time has been severely constrained lately.

choldgraf:
Pssshhh next you're gonna tell me that a global crisis has happened or something 😅😭

SylvainCorlay:
Should we move the mechanism we have in Voilà to handle that to nbclient?

The `VoilaExecutor` inerits from `NotebookClient`, and includes a mock `OutputWidget` which mimicks the behavior of the real front-end output widget.

This was developed by @maartenbreddels and I think he should be credited with the commit should this be included in nbclient.

choldgraf:
I was thinking the same thing actually...I think I'd be +1 on it unless somebody else has a cleaner solution in mind

SylvainCorlay:
I think this solution is actually clean.

SylvainCorlay:
We may want to provide a hook for providers of (other) comm targets than jupyter-widgets implementing things that impact the notebook state to provide similar pure python "front ends" like this output widget.

Will open a PR for this tomorrow.

choldgraf:
That's sounds great, I didn't mean to imply this wasn't a clean solution, it seems good to me. I will defer to you all about the right approach 👍

MSeal:
I'm all for adding this from Voilà. While I didn't get around to implementing the same logic I'll gladly review a PR

maartenbreddels:
I agree, I think nbclient is the right place for this. Happy to port the work + test to nbclient.
Long term I think the way the Output widget works should change.

MSeal:
With the 0.4.0 release this is now solved. Note you still have to "Trust Notebook" in classic UI when loading in a browser after execution like other js generating components.. Thanks @maartenbreddels for tackling the implementation!

choldgraf:
Indeed, thanks so much @maartenbreddels :-)

Issue #58, opened 2020-05-17
Errors are upgrading to 0.3.0


palewire:
Some users following a pattern similar to the one described in #48 are getting new errors are upgrading to 0.3. When they downgrade to 0.2, the errors are gone. Here's a sample traceback.

```python
Traceback (most recent call last):
  File "update.py", line 154, in <module>
    cli()
  File "/Users/apesce/.local/share/virtualenvs/coronavirus-tracker-dljAgG8F/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/apesce/.local/share/virtualenvs/coronavirus-tracker-dljAgG8F/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/apesce/.local/share/virtualenvs/coronavirus-tracker-dljAgG8F/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/apesce/.local/share/virtualenvs/coronavirus-tracker-dljAgG8F/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/apesce/.local/share/virtualenvs/coronavirus-tracker-dljAgG8F/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "update.py", line 85, in process
    _execute_notebook(
  File "update.py", line 29, in _execute_notebook
    client.execute()
  File "/Users/apesce/.local/share/virtualenvs/coronavirus-tracker-dljAgG8F/lib/python3.8/site-packages/nbclient/util.py", line 37, in wrapped
    result = loop.run_until_complete(coro(self, *args, **kwargs))
  File "/usr/local/Cellar/python@3.8/3.8.1/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 612, in run_until_complete
    return future.result()
  File "/Users/apesce/.local/share/virtualenvs/coronavirus-tracker-dljAgG8F/lib/python3.8/site-packages/nbclient/client.py", line 471, in async_execute
    async with self.async_setup_kernel(**kwargs):
[...]

PR #68, opened 2020-05-26
feat: implement Output widget that mimics a frontend

maartenbreddels:
This is a port of https://github.com/voila-dashboards/voila/pull/91 and subsequent fixes.

### DIFF
--- a/nbclient/_version.py
+++ b/nbclient/_version.py
@@ -1,1 +1,1 @@
-version = '0.3.1'
+version = '0.4.0-dev.0'
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -306,6 +306,10 @@ class NotebookClient(LoggingConfigurable):
                 'OutputModel': OutputWidget
             }
         }
+        # comm_open_handlers should return an object with a .handle_msg(msg) method or None
+        self.comm_open_handlers = {
+            'jupyter.widget': self.on_comm_open_jupyter_widget
+        }
 
     def reset_execution_trackers(self):
         """Resets any per-execution trackers.
@@ -318,7 +322,7 @@ class NotebookClient(LoggingConfigurable):
         # to support nested use of output widgets.
         self.output_hook_stack = collections.defaultdict(list)
         # our front-end mimicing Output widgets
-        self.widget_objects = {}
+        self.comm_objects = {}
 
     def start_kernel_manager(self):
         """Creates a new kernel manager.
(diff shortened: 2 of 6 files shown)


maartenbreddels:
Fixes #24

I've added a changelog entry and bumped the version number. It's a new workflow I'm using, which makes it easier to track changes (and keep semantic versioning).
Happy to remove it if people don't like it though.

maartenbreddels:
Failure at https://travis-ci.org/github/jupyter/nbclient/jobs/691227550 seems unrelated, but a bit worrying.

davidbrochart:
This test fails randomly, I know it's a problem. I'm restarting it.

davidbrochart:
Now it passes :smile:

nbclient/_version.py
```diff
@@ -1 +1 @@
-version = '0.3.1'
+version = '0.3.2-dev.0'
```
SylvainCorlay, line 1: I think this should be 0.4.0-dev.0 because this will break Voilà, which inherits from the executor, which may duplicate the whole Output widget logic.

nbclient/client.py
```diff
@@ -832,6 +856,24 @@ def handle_comm_msg(self, outs, msg, cell_index):
             self.widget_state.setdefault(content['comm_id'], {}).update(data['state'])
             if 'buffer_paths' in data and data['buffer_paths']:
                 self.widget_buffers[content['comm_id']] = self._get_buffer_data(msg)
+        # There are cases where we need to mimic a frontend, to get similar behaviour as
+        # when using the Output widget from Jupyter lab/notebook
+        if msg['msg_type'] == 'comm_open' and msg['content'].get('target_name') == 'jupyter.widget':
```
SylvainCorlay, line 861: Maybe we need to register comm_open and comm_msg handlers per comm target, eventually.

So instead of doing `msg['content'].get('target_name') == 'jupyter.widget'` we would do `msg['content'].get('target_name') in comm_open_handlers`.

nbclient/client.py
```diff
@@ -832,6 +856,24 @@ def handle_comm_msg(self, outs, msg, cell_index):
...
+            content = msg['content']
+            data = content['data']
+            state = data['state']
+            comm_id = msg['content']['comm_id']
+            if state['_model_module'] == '@jupyter-widgets/output' and\
```
SylvainCorlay, line 866: Same here, probably filter by target name.

nbclient/_version.py
maartenbreddels: I agree, it changes behavior.

maartenbreddels:
docs fail because of missing ipykernel dependency:
```ModuleNotFoundError: No module named 'ipykernel'```

Not sure how to handle this, we probably don't want or need to depend on it right?

It is only needed for `from ipykernel.jsonutil import json_clean` Maybe some other package has this as well.

**Comment by davidbrochart** [comment]
I think @choldgraf should be able to help us here.

File: nbclient/output_widget.py [hunk 4]
```diff
@@ -0,0 +1,85 @@
+from ipykernel.jsonutil import json_clean
```
SylvainCorlay, line 1: Indeed, this seems like a weak reason to depend on ipykernel.

Should we duplicate the implementation of `json_clean` in this package instead?

SylvainCorlay: There is a jsonutils in `jupyter_client` as well but without that function.

How about adding a jsonutils here?

maartenbreddels:
I added and blackified the jsontil.py, we may want to find a better long term solution some day.

SylvainCorlay:
Yes! Really happy to see things go through.

SylvainCorlay:
@choldgraf would you be fine with releasing 0.4.0 now?

nbclient/jsonutil.py
```diff
@@ -0,0 +1,205 @@
...
+
+    if isinstance(obj, atomic_ok):
+        return obj
+
+    if isinstance(obj, bytes):
+        if py3compat.PY3:
+            # unanmbiguous binary data is base64-encoded
+            # (this probably should have happened upstream)
+            return b2a_base64(obj).decode('ascii')
+        else:
```
MSeal, line 167: Don't need this code branch

MSeal:
> I added and blackified the jsontil.py, we may want to find a better long term solution some day.

Yes that did seem like overkill, but if it's coming from a working solution we can use what works until someone refactors later.

> @choldgraf would you be fine with releasing 0.4.0 now?

Let's double check what's still open in issues / PRs first. I added you as a maintainer on PyPI @SylvainCorlay

choldgraf:
this is super exciting :-)  thanks @maartenbreddels and @SylvainCorlay for pushing this through!

choldgraf:
I am +1 on pushing out a new release unless there's an obvious blocker in our issues or a super low-hanging fruit

MSeal:
I think https://github.com/jupyter/nbclient/pull/64 and maybe https://github.com/jupyter/nbclient/issues/58 could be addressed before a release. We're still in sub 1.0 releases so it's also ok for us to ship now and prep a 0.5 for those other two.

On that note we should set a goal for 1.0 release -- it's probably pretty close? I'll start an issue to track what we want to target to be considered feature complete

PR #70, opened 2020-05-27
Automatically use nest_asyncio if the event loop is already running

davidbrochart:
This PR removes the `nest_asyncio` trait from the `NotebookClient` class, and thus the keyword argument `nest_asyncio` for e.g. the `execute()` method.


davidbrochart:
I think we should close it in favor of #71.

SylvainCorlay:
OK, closing as discussed.

PR #71, opened 2020-05-27
fix: always run a coroutine and work with tornado

maartenbreddels:
This is a port of what I use in [vaex](https://github.com/vaexio/vaex/blob/10e1d5077021a0590d3750d62d4d56a27a84af82/packages/vaex-core/vaex/asyncio.py)

This has been running for quite a while on CI, and seems pretty solid. If an event loop is present  it will patch it with nested-asyncio and will also patch tornado, which should solve:
 * https://github.com/nteract/papermill/issues/490

Alternative to https://github.com/jupyter/nbclient/pull/67

See also https://github.com/erdewit/nest_asyncio/issues/23 and https://github.com/tornadoweb/tornado/issues/2753

This also checks to see if it is being run in an IPython context (I'm not sure the check makes sense, I just check if it is imported), and will give an error if not v7+, otherwise cells in a notebook may run out of order (a known? issue with older versions of IPython I believe), see also https://github.com/erdewit/nest_asyncio/issues/9

### DIFF
--- a/nbclient/tests/util.py
+++ b/nbclient/tests/util.py
@@ -1,0 +1,57 @@
+import asyncio
+
+import tornado
+
+from nbclient.util import run_sync
+
+
+@run_sync
+async def some_async_function():
+    await asyncio.sleep(0.01)
+    return 42
+
+
+def test_nested_asyncio_with_existing_ioloop():
+    ioloop = asyncio.new_event_loop()
+    try:
+        asyncio.set_event_loop(ioloop)
+        assert some_async_function() == 42
+        assert asyncio.get_event_loop() is ioloop
+    finally:
+        asyncio._set_running_loop(None)  # it seems nest_asyncio doesn't reset this
+
+
+def test_nested_asyncio_with_no_ioloop():
+    asyncio.set_event_loop(None)
+    try:
+        assert some_async_function() == 42
+    finally:
+        asyncio._set_running_loop(None)  # it seems nest_asyncio doesn't reset this
+
+
+def test_nested_asyncio_with_tornado():
+    # This tests if tornado accepts the pure-Python Futures, see
+    # https://github.com/tornadoweb/tornado/issues/2753
+    # https://github.com/erdewit/nest_asyncio/issues/23
+    asyncio.set_event_loop(asyncio.new_event_loop())
+    ioloop = tornado.ioloop.IOLoop.current()
+
+    async def some_async_function():
+        future = asyncio.ensure_future(asyncio.sleep(0.1))
+        # this future is a different future after nested-asyncio has patched
+        # the asyncio module, check if tornado likes it:
+        ioloop.add_future(future, lambda f: f.result())
+        await future
+        return 42
+
+    def some_sync_function():
+        return run_sync(some_async_function)()
+
+    async def run():
+        # calling some_async_function directly should work
+        assert await some_async_function() == 42
+        # but via a sync function (using nested-asyncio) can lead to issues:
+        # https://github.com/tornadoweb/tornado/issues/2753
+        assert some_sync_function() == 42
+
+    ioloop.run_sync(run)
--- a/nbclient/util.py
+++ b/nbclient/util.py
@@ -4,7 +4,51 @@
 # Distributed under the terms of the Modified BSD License.
 
 import asyncio
+import sys
 import inspect
+
+
+def check_ipython():
+    # original from vaex/asyncio.py
+    IPython = sys.modules.get('IPython')
+    if IPython:
+        IPython_version = tuple(map(int, IPython.__version__.split('.')))
+        if IPython_version < (7, 0, 0):
+            raise RuntimeError(f'You are using IPython {IPython.__version__} while we require'
+                               '7.0.0+, please update IPython')
+
+
+def check_patch_tornado():
+    """If tornado is imported, add the patched asyncio.Future to its tuple of acceptable Futures"""
+    # original from vaex/asyncio.py
+    if 'tornado' in sys.modules:
+        import tornado.concurrent
+        if asyncio.Future not in tornado.concurrent.FUTURES:
+            tornado.concurrent.FUTURES = tornado.concurrent.FUTURES + (asyncio.Future, )
+
+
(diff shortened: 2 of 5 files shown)


nbclient/util.py
```diff
@@ -4,9 +4,52 @@
...
+        loop = None
+        had_loop = False
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+    try:
+        if had_loop:
+            check_ipython()
+            import nest_asyncio
+            nest_asyncio.apply()
+            check_patch_tornado()
```
davidbrochart, line 46: We might want to do that only once if we had a loop, not at each coroutine execution?

davidbrochart:
Awesome work @maartenbreddels!

maartenbreddels:
My problem with this PR is that I don't have a test, I have trouble reproducing the original issue. So lets put it to draft first.

nbclient/util.py
```diff
@@ -4,9 +4,52 @@
...
+
+
+def just_run(coro):
+    """Make the coroutine run, even if there is an event loop running (using nest_asyncio)"""
+    # original from vaex/asyncio.py
+    try:
+        loop = asyncio.get_event_loop()
+        had_loop = True
+    except RuntimeError:
+        loop = None
```
davidbrochart, line 37: Not needed.

nbclient/util.py
maartenbreddels: Well, the patch is not applied when tornado is not imported (this happens in vaex, where I first call this function, then start the server, and call this function again). Also the patch is only applied once (it checks if the Future is in the set).

nbclient/util.py
```diff
@@ -4,9 +4,52 @@
...
+    try:
+        if had_loop:
+            check_ipython()
+            import nest_asyncio
+            nest_asyncio.apply()
+            check_patch_tornado()
+        return loop.run_until_complete(coro)
+    finally:
+        if not had_loop:  # remove loop if we did not have one
+            asyncio.set_event_loop(None)
```
davidbrochart, line 50: Why not keep the event loop we just created?

nbclient/util.py
davidbrochart: OK, but what about `check_ipython()` and `nest_asyncio.apply()`?

nbclient/util.py
maartenbreddels: To keep things as they were, otherwise you leave a non-running loop.

davidbrochart: On the other hand we are probably going to recreate an event loop in a subsequent call. I don't think having an idle event loop is a problem?

nbclient/util.py
maartenbreddels: For check_ipython the same holds (it can be imported later on), and nest_asyncio has a check build in, in the end, simplicity, otherwise we need to check things that nest_asyncio already checks.

nbclient/util.py
maartenbreddels: I *think* we need to check if the existing loop is then non-running, and make it run again? I thought this would be safer, a clean start each time, e.g. no old stuff leftover because of obscure bugs.

Also, this solution has survived quite a while in vaex' CI, so I would be reluctant to change this unless we have good test coverage, which would make me really happy :)

So not against it, just being conservative here :)

nbclient/util.py
```diff
@@ -4,9 +4,52 @@
...
+        import tornado.concurrent
+        if asyncio.Future not in tornado.concurrent.FUTURES:
+            tornado.concurrent.FUTURES = tornado.concurrent.FUTURES + (asyncio.Future, )
+
+
+def just_run(coro):
+    """Make the coroutine run, even if there is an event loop running (using nest_asyncio)"""
+    # original from vaex/asyncio.py
+    try:
+        loop = asyncio.get_event_loop()
```
davidbrochart, line 34: `asyncio.get_event_loop()` will create an event loop if no one already exists, so can we say we had a loop?

nbclient/util.py
davidbrochart: :+1:

nbclient/util.py
maartenbreddels: I thought it did not? But please double check

https://docs.python.org/3.6/library/asyncio-eventloops.html#asyncio.AbstractEventLoopPolicy.get_event_loop

`Raises an exception in case no event loop has been set`

davidbrochart: That's python3.6, now we have: https://docs.python.org/3/library/asyncio-eventloop.html

maartenbreddels: so... whow :) That's kind of a breaking change isn't it... any idea what the best behavior should be?

davidbrochart: Actually I'm not sure it has changed. But in any case, we could try and run the coroutine and if we catch an `event loop already running` exception, then we had a loop.

maartenbreddels: It depends on the policy it seems: https://docs.python.org/3.6/library/asyncio-eventloops.html#asyncio.AbstractEventLoopPolicy.get_event_loop

davidbrochart: Let's say we support only the default policy.

maartenbreddels: I think you are right that the only way to detect a running loop in py36 is to just try :(

maartenbreddels:
Using the private API `asyncio._get_running_loop()` is the most portable way it seems to find out of we have a running ioloop.

davidbrochart:
Thanks a lot for the tests and the awesome work!

nbclient/tests/util.py
```diff
@@ -0,0 +1,58 @@
...
+    return 42
+
+
+def test_nested_asyncio_with_existing_ioloop():
+    ioloop = asyncio.new_event_loop()
+    try:
+        asyncio.set_event_loop(ioloop)
+        assert some_async_function() == 42
+        assert asyncio.get_event_loop() is ioloop
+        print(ioloop, id(ioloop))
```
davidbrochart, line 20: A leftover?

MSeal:
0.4.0a0 is released fyi

PR #74, opened 2020-05-29
fix: if a comm target is not know, give a warning and continue

maartenbreddels:
This was the reason https://github.com/voila-dashboards/voila/pull/621 failed, since that was using the matplotlib comm.

We now give a warning:
```bash
$ python -m nbconvert nbclient/tests/files/Other\ Comms.ipynb --execute
.....
[NbConvertApp] WARNING | No handler found for comm target 'this-comm-tests-a-missing-handler'
```

and just continue executing.

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -853,11 +853,15 @@ class NotebookClient(LoggingConfigurable):
         # There are cases where we need to mimic a frontend, to get similar behaviour as
         # when using the Output widget from Jupyter lab/notebook
         if msg['msg_type'] == 'comm_open':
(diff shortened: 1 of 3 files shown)


maartenbreddels:
I guess the failure is unrelated

davidbrochart:
Yes, again `test_many_parallel_notebooks`... we should do something about it.

MSeal:
Yeah I thought we'd worked out all the race conditions around forked pyzmq by isolating the contexts but I still do see occasional hangs in the tests. I don't hit this outside of travis tests so it's made it hard to debug. @minrk fyi since we were chatting about fork-safe pyzmq.

PR #75, opened 2020-05-29
feat: support nbagg matplotlib backend (WIP)

maartenbreddels:
This would solve https://github.com/jupyter/nbconvert/issues/1081 and would also make the old legacy nbagg backend work with voila (at least render it).

However, I cannot get it to work, the backend does not send me back the image, which then should be injected in the div.

Note: if you read this, consider using ipympl instead https://github.com/matplotlib/ipympl


maartenbreddels:
It now fails for the right reason, which is that the mime bundle only contains a div, while it should contain an image with the base64 encoded data. This is not implemented because I do not get the message.

maartenbreddels:
@tacaswell @martinRenou
Maybe you have ideas on how to fix this? The protocol is a bit hard to digest.

MSeal:
Thanks for opening the WIP

Issue #73, opened 2020-05-27
Master test failure


SylvainCorlay:
Thee tests on master appear to be failing on Windows + Python 3.6 over `OSError: [WinError 6] The handle is invalid`.

I re-started the tests to see if this occurs all the time, because the PR that was merged last was passing.

SylvainCorlay:
Edit: always failing on windows-latest 3.6.

davidbrochart:
It looks a lot like the issue we had in [jupyter_server](https://github.com/jupyter/jupyter_server/pull/191#issuecomment-599634432), basically there was a bug in Python on Windows that has been fixed and back-ported to python >= 3.7 (https://bugs.python.org/issue37380).

MSeal:
Thanks for raising. If someone hasn't dug into fixing this in a couple days I should have time at the end of the week to fix builds. There's a number of doc build failures across projects as well that need tweaks to repair :/

willingc:
nbclient doc builds are passing.

SylvainCorlay:
> has been fixed and back-ported to python >= 3.7 (https://bugs.python.org/issue37380).

Thanks! Do you know if there is an advertized workaround so that we don't have to drop support for windows and Python 3.6.

davidbrochart:
I couldn't find any.

SylvainCorlay:
We should ask for one on the CPython issue.

davidbrochart:
See #77

MSeal:
Looks like it's all resolved now. Thanks @davidbrochart !

PR #77, opened 2020-05-30
Fix for windows/python<3.7

davidbrochart:
This patches python's `subprocess` module as it is done in https://github.com/python/cpython/pull/15706.
We suspect #73 is due to https://bugs.python.org/issue37380, but the fix was not back-ported to python<3.7.


davidbrochart:
This seems to solve the `OSError: [WinError 6] The handle is invalid` error.

MSeal:
Thanks @davidbrochart !

PR #80, opened 2020-06-01
Release 0.4.0a1

davidbrochart:


MSeal:
Thanks for updating that! If things look stable we should consider flipping this to the full 0.4.0 release.

PR #83, opened 2020-06-19
Type checking

davidbrochart:


choldgraf:
This LGTM but I'm probably not the best to review since I am still a heathen that has never used Python type checking before :-)

davidbrochart:
Same for me just before this PR :smile:

nbclient/client.py
```diff
@@ -51,7 +55,7 @@ class NotebookClient(LoggingConfigurable):
         ),
     ).tag(config=True)
 
-    timeout_func = Any(
+    timeout_func: t.Any = Any(
```
MSeal, line 58: Usually you do from `typing import Any`, but this is fine

nbclient/client.py
```diff
@@ -273,7 +277,11 @@ def _kernel_manager_class_default(self):
         )
     )
 
-    def __init__(self, nb, km=None, **kw):
+    def __init__(
+            self,
+            nb: NotebookNode,
+            km: t.Optional[KernelManager] = None,
```
MSeal, line 283: Technically there's a KernelManagerABC.. but I don't *think* anyone directly inherits from the abc other than KernelManager. But it might be a good idea to change to over.

nbclient/client.py
```diff
@@ -344,13 +354,13 @@ async def _async_cleanup_kernel(self):
         finally:
             # Remove any state left over even if we failed to stop the kernel
             await ensure_async(self.km.cleanup())
-            if getattr(self, "kc"):
+            if self.kc is not None:
```
MSeal, line 357: I believe this one actually needs to be `getattr(self, "kc")` or `getattr(self, "kc") is not None` because the KernelGateway (and possibly other?) code paths sometimes don't assign any kc to the object by the time it reaches this call sometimes. I see you were testing this with asserts within the functions in places, but I can't quite recall the exact circumstances that cause the issue originally.

nbclient/client.py
```diff
@@ -419,7 +430,7 @@ def setup_kernel(self, **kwargs):
 
         # Can't use run_until_complete on an asynccontextmanager function :(
         if self.km is None:
-            self.start_kernel_manager()
+            self.km = self.start_kernel_manager()
```
MSeal, line 433: Not really necessary to do the double assignment since `self.start_kernel_manager()` assigns self.km regardless, but doesn't really hurt anything either for protecting subclasses not implementing the same side-effect.

nbclient/client.py
```diff
@@ -607,22 +642,33 @@ def _get_timeout(self, cell):
...
             self.log.error("Kernel died while waiting for execute reply.")
             raise DeadKernelError("Kernel died")
 
-    async def async_wait_for_reply(self, msg_id, cell=None):
+    async def async_wait_for_reply(
+            self,
+            msg_id: str,
+            cell: t.Optional[NotebookNode] = None) -> t.Optional[t.Dict]:
+
+        assert self.kc is not None
```
MSeal, line 671: A little odd to keep the asserts in non-test code.

nbclient/client.py
```diff
@@ -820,13 +879,13 @@ def output(self, outs, msg, display_id, cell_index):
             # default output behaviour (e.g. OutputWidget)
             hook = self.output_hook_stack[parent_msg_id][-1]
             hook.output(outs, msg, display_id, cell_index)
-            return
+            return None
```
MSeal, line 882: Had to reread the PEP on this, but when returning elsewhere with a value it is recommended to have a `return None` explicitly where exiting early without an object. So good catch and I'll have to remember that in the future!

choldgraf: explicit is better than implicit! :-)

nbclient/client.py
MSeal: @kevin-bates FYI I think we should shift the RemoteKernelManager to inherit from the KernelManagerABC so we can have a consistent type check / interface. From looking at the class I believe this wouldn't disrupt anything therein.

kevin-bates: If RemoteKernelManager inherits directly from KernelManagerABC instead of the concrete KernelManager, how does it then leverage the implementation of KernelManager? 

RMK is not an independent implementation of KernelManagerABC - otherwise that's what it would inherit from.  It needs to be plugged into the framework and that is accomplished by deriving from KernelManager.

MSeal: I tracked down the inheritance and it ends up at [MappingKernelManager](https://github.com/jupyter/notebook/blob/6e9256b0641a85baf664e846515085bac2c082a3/notebook/services/kernels/kernelmanager.py#L42)->[MultiKernelManager](https://github.com/jupyter/jupyter_client/blob/15b3b0566d962bbc042d21b858caeca2e870209a/jupyter_client/multikernelmanager.py#L43) -- which means we should instead make `MultiKernelManager` inherit the `KernelManagerABC` (OR make this line take a union of managers?) for type hinting check purposes. Drifting from the topic at hand (can move elsewhere if need be) is there a historical reason the `MappingKernelManager` is in notebook over jupyter_client? Where do we draw the line on custom extensions within the jupyter github repos for kernel managers?

> RKM is not an independent implementation of KernelManagerABC - otherwise that's what it would inherit from. It needs to be plugged into the framework and that is accomplished by deriving from KernelManager.

Think of KernelManagerABC as an interface mixin -- you can inherit from it and other things without issue using [Python's default inheritance ordering](https://data-flair.training/blogs/python-multiple-inheritance) for overlapping method names. Given the chain of existing inheritance it does reach into the jupyter_client as I pointed out above, so likely we need to just resolve that in jupyter_client without wanting/needing any changes in EG.

nbclient/client.py
davidbrochart: Yes, this is enforced buy the type checker that complains with `error: Return value expected` otherwise.

nbclient/client.py
davidbrochart: Here too, with the assert we get `error: Item "None" of "Optional[Any]" has no attribute "shell_channel"` later in the method, because the type checker has no idea `self.kc` has a non-`None` value at that time.

nbclient/client.py
davidbrochart: If we don't assign to `self.km` we get `error: Item "None" of "Optional[Any]" has no attribute "has_kernel"` with the following line.

nbclient/client.py
davidbrochart: So I changed it because otherwise we get `error: Item "None" of "Optional[Any]" has no attribute "stop_channels"` with the following line. I think it should be:
```python
if getattr(self, "kc") and self.kc is not None:
```

nbclient/client.py
davidbrochart: Yes but some names in `typing` collide with `traitlets` (`List`, `Any`, `Dict`), it's annoying.

MSeal: Oh... yeah good point

nbclient/client.py
MSeal: sounds good to me

nbclient/client.py
```diff
@@ -642,26 +688,22 @@ async def async_wait_for_reply(self, msg_id, cell=None):
...
     _wait_for_reply = wait_for_reply
 
-    def _timeout_with_deadline(self, timeout, deadline):
-        if deadline is not None and deadline - monotonic() < timeout:
-            timeout = deadline - monotonic()
-
-        if timeout < 0:
-            timeout = 0
-
-        return timeout
```
davidbrochart, line 657: Note that this method seemed to be never called, so I removed it.

davidbrochart, line 657: Note that this method seemed to be never called, so I removed it.

MSeal: Huh, wonder how that happened. Probably we missed nuking it in an earlier refactor. Good catch.

nbclient/client.py
MSeal: Let's merge for now and we can open new issues if there's problems with EG patterns in combination.

MSeal:
Thanks @davidbrochart for knocking this out!

PR #86, opened 2020-06-27
DOC: A few reST syntactic changes

mgeier:
Also enables the `"any"` role by default, which automatically creates links and -- very importantly -- creates warnings on broken links.

BTW, this also checks for warnings on CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,7 @@ jobs:
       - run:
           name: Build docs to store
           command: |
-            sphinx-build -b html docs/ docs/_build/html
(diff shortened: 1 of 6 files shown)


mgeier:
I don't know whether `tests.test_client.IPY_MAJOR` is needed. The tests seem to pass without it.

If it actually *is* needed, I'll have to mock `IPython` more thoroughly or add it to the doc dependencies.

MSeal:
I'm going to go ahead and merge. If there's more follow-up changes I can review those quickly as well.

PR #87, opened 2020-06-30
Fix usage of hasattr for kc

JoostvDoorn:
Usage of `getattr` was wrong here, replaced with `hasattr`


MSeal:
I believe the test failures are related to the socket cleanup fixes in jupyter_client and nbclient tests not properly respecting cleanups. I'll look into it later and make sure it's that and not related to this PR later.

JoostvDoorn:
On second inspection I mistakenly thought this issue was not yet fixed, it has already been fixed. In #83 `self.kc = None` was added.

PR #90, opened 2020-07-09
Check if kernel is alive during cell execution

### LINKED ISSUE #88: Nbclient waits indefinitely on killed subprocesses
We use nbclient with papermill to run notebooks in docker containers. If a notebook exceed the allocated memory (of the container) only the notebook subprocess seems to be killed (or dies), this seems to be not detected by nbclient, resulting in an infinite running process.

```python
import nbformat
from nbclient import NotebookClient

print("Start")
with open("test_notebook.ipynb") as f:
    nb = nbformat.read(f, as_version=4)

client = NotebookClient(nb, kernel_name="python3")
client.execute()

print("Done")
```

This issue can be reproduced in two ways, with the script above:

- Running a python debuger, setting a breakpoint here https://github.com/jupyter/nbclient/blob/master/nbclient/client.py#L521 and killing the subprocess by hand using the system monitor, and then continuing execution. _Note: This can create the same type of issue (stuck in an infinite loop), but the solution might not completely overlap with the issue we are experiencing._
- Or using the following steps with docker. :

<details>

First ensure that "cgroup swap limit capabilities" are enabled (if you have a lot of swap space). I used Ubuntu. See here for more info:
https://docs.docker.com/engine/install/linux-postinstall/#your-kernel-does-not-support-cgroup-swap-limit-capabilities

Create the following files

Dockerfile:
```docker
FROM python:3
WORKDIR /usr/src/app
RUN pip install numpy ipykernel nbclient
COPY . .
CMD [ "python", "./test.py" ]
```

test_notebook.ipynb:
```python
import numpy as np

test = np.random.random((110, 10000000))
```

Build and run docker container (this will result in an infinite running process):

```
docker build -t my-python-app .
docker run -it --memory-swap 1000m --memory 1000m --rm --name my-running-app my-python-app
```

Run docker container without any issues:
```
docker run -it --rm --name my-running-app my-python-app
```
</details>

The issue may be caused by a missing timeout on the following line, but adding a timeout does not work for me:
https://github.com/jupyter/nbclient/blob/4314a44b3074cfedd4d05128d81765b285194ab8/nbclient/client.py#L627

davidbrochart:
Fixes #88

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -644,6 +644,7 @@ class NotebookClient(LoggingConfigurable):
             try:
                 await self._async_check_alive()
             except DeadKernelError:
+                assert self.task_poll_for_reply is not None
                 self.task_poll_for_reply.cancel()
                 return
 
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -358,6 +358,7 @@ class NotebookClient(LoggingConfigurable):
             if getattr(self, "kc") and self.kc is not None:
                 await ensure_async(self.kc.stop_channels())
                 self.kc = None
+                self.km = None
 
     _cleanup_kernel = run_sync(_async_cleanup_kernel)
 
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -310,6 +310,7 @@ class NotebookClient(LoggingConfigurable):
     def reset_execution_trackers(self) -> None:
         """Resets any per-execution trackers.
         """
+        self.task_poll_for_reply: t.Optional[asyncio.Future] = None
         self.code_cells_executed = 0
         self._display_id_map = {}
         self.widget_state: t.Dict[str, t.Dict] = {}
@@ -584,7 +585,8 @@ class NotebookClient(LoggingConfigurable):
             msg_id: str,
             cell: NotebookNode,
             timeout: t.Optional[int],
-            task_poll_output_msg: asyncio.Future) -> t.Dict:
+            task_poll_output_msg: asyncio.Future,
+            task_poll_kernel_alive: asyncio.Future) -> t.Dict:
 
         assert self.kc is not None
         new_timeout: t.Optional[float] = None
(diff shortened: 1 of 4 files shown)


davidbrochart:
I don't know about the failures, they are related to pyzmq and were already present in the [last commit](https://github.com/jupyter/nbclient/commit/e1c0ba7c5dcc3547784ac47fba65c8cfd2e16722).

davidbrochart:
The failing test is caused by the deprecation warning, I opened an issue in https://github.com/jupyter/jupyter_client/issues/558.

davidbrochart:
Looks like we got bitten by https://github.com/jupyter/jupyter_client/issues/487:
```
E         + Exception in thread Thread-3:
E         + Traceback (most recent call last):
E         +   File "/opt/hostedtoolcache/Python/3.8.3/x64/lib/python3.8/threading.py", line 932, in _bootstrap_inner
E         +     self.run()
E         +   File "/home/runner/work/nbclient/nbclient/.tox/py38/lib/python3.8/site-packages/ipykernel/heartbeat.py", line 93, in run
E         +     self._bind_socket()
E         +   File "/home/runner/work/nbclient/nbclient/.tox/py38/lib/python3.8/site-packages/ipykernel/heartbeat.py", line 74, in _bind_socket
E         +     self._try_bind_socket()
E         +   File "/home/runner/work/nbclient/nbclient/.tox/py38/lib/python3.8/site-packages/ipykernel/heartbeat.py", line 61, in _try_bind_socket
E         +     return self.socket.bind('%s://%s' % (self.transport, self.ip) + c + str(self.port))
E         +   File "zmq/backend/cython/socket.pyx", line 550, in zmq.backend.cython.socket.Socket.bind
E         +   File "zmq/backend/cython/checkrc.pxd", line 26, in zmq.backend.cython.checkrc._check_rc
E         + zmq.error.ZMQError: Address already in use
```
Relaunching tests.

davidbrochart:
@MSeal I will be AFK for a few days, I let you squash and merge if you think this is fine.

MSeal:
Thanks so much for getting this fixed. I'll kick off a patch release.

choldgraf:
wahoo! 🥇

PR #93, opened 2020-07-28
cleanup async_start_new_kernel_client

### LINKED ISSUE #91: Exception in async_setup_kernel
I am getting the following error in nbclient when running voila 0.2 alpha with the classic template

```
  File "voila/handler.py", line 126, in _jinja_notebook_execute
    result = await self.executor.async_execute()
  File "nbclient/client.py", line 520, in async_execute
    async with self.async_setup_kernel(**kwargs):
  File "async_generator/_util.py", line 34, in __aenter__
    return await self._agen.asend(None)
  File "nbclient/client.py", line 478, in async_setup_kernel
    if not self.km.has_kernel:
AttributeError: 'AsyncMappingKernelManager' object has no attribute 'has_kernel'
```

SylvainCorlay:
Fixes #91 

This makes it clear that the `km` attribute should *always* be a `KernelManager` and not a `MultiKernelManager`. These two classes have different APIs and behave significantly differently and we should not have handled the case of a `MultiKernelManager` in this function.

I would be in favor of calling it a 0.5.0 when this gets in.

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -362,7 +362,7 @@ class NotebookClient(LoggingConfigurable):
 
     _cleanup_kernel = run_sync(_async_cleanup_kernel)
 
-    async def async_start_new_kernel_client(self, **kwargs) -> t.Tuple[KernelClient, str]:
+    async def async_start_new_kernel_client(self, **kwargs) -> KernelClient:
         """Creates a new kernel client.
 
         Parameters
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -384,29 +384,12 @@ class NotebookClient(LoggingConfigurable):
         if resource_path and 'cwd' not in kwargs:
             kwargs["cwd"] = resource_path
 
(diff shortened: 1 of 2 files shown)


nbclient/client.py
```diff
@@ -384,37 +384,20 @@ async def async_start_new_kernel_client(self, **kwargs) -> t.Tuple[KernelClient,
...
+        self.kc = self.km.client()
         await ensure_async(self.kc.start_channels())
         try:
             await ensure_async(self.kc.wait_for_ready(timeout=self.startup_timeout))
         except RuntimeError:
             await self._async_cleanup_kernel()
             raise
         self.kc.allow_stdin = False
-        return self.kc, kernel_id
+        return self.kc
```
davidbrochart, line 400: Now that we return only `self.kc`, we should change the return type on this method to `KernelClient`.

davidbrochart:
Thanks @SylvainCorlay for cleaning this part. It was a mistake for me to try and support a `KernelManager` and a `MultiKernelManager` in `async_start_new_kernel_client`.

Issue #92, opened 2020-07-28
Rename start_kernel_manager into create_kernel_manager


SylvainCorlay:
The former makes it look like it is starting a kernel or something like this, which it is just instantiating the kernel manager class.

davidbrochart:
Agreed, and that would make it consistent with the docstring:
https://github.com/jupyter/nbclient/blob/bf77f5ac3f3349c91d4fca40b2b909f254379fb7/nbclient/client.py#L325

davidbrochart:
Fixed in #94

PR #94, opened 2020-07-30
Split start_new_kernel_client into kernel and client creation

davidbrochart:
Separates `(async_)start_new_kernel_client` into:
- `(async_)start_new_kernel`: starts a new kernel.
- `(async_)start_new_kernel_client`: starts a new client of the kernel.

This gives more flexibility as to creating the kernel and/or the client from outside. Internally, both methods are called so it doesn't change anything.

Also in this PR:
- `start_kernel_manager` renamed to `create_kernel_manager` (see #92).
- `self.owns_km` indicates if the kernel manager is created by the `NotebookClient` or if it is passed from outside. If the `NotebookClient` owns the kernel manager, the cleanup of the kernel (client and manager) will be done by default. If not, the cleanup won't be done by default (previously, it was always done by default).

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -295,6 +295,7 @@ class NotebookClient(LoggingConfigurable):
         super().__init__(**kw)
         self.nb: NotebookNode = nb
         self.km: t.Optional[KernelManager] = km
+        self.owns_km: bool = km is None  # whether the NotebookClient owns the kernel manager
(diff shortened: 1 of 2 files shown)


MSeal:
I'll wait for Sylvain to take a look to merge

MSeal:
Going to merge since it's been a bit without other input

PR #96, opened 2020-08-05
Check language_info key in kernel info received message content

### LINKED ISSUE #95: info_msg contains no language_info
In https://readthedocs.org/projects/myst-nb/builds/11597535/,
I'm getting the error:

```pytb
  File "/home/docs/checkouts/readthedocs.org/user_builds/myst-nb/envs/latest/lib/python3.7/site-packages/jupyter_cache/executors/basic.py", line 166, in execute
    record_timing=False,
  File "/home/docs/checkouts/readthedocs.org/user_builds/myst-nb/envs/latest/lib/python3.7/site-packages/nbclient/client.py", line 1077, in execute
    return NotebookClient(nb=nb, resources=resources, km=km, **kwargs).execute()
  File "/home/docs/checkouts/readthedocs.org/user_builds/myst-nb/envs/latest/lib/python3.7/site-packages/nbclient/util.py", line 74, in wrapped
    return just_run(coro(*args, **kwargs))
  File "/home/docs/checkouts/readthedocs.org/user_builds/myst-nb/envs/latest/lib/python3.7/site-packages/nbclient/util.py", line 53, in just_run
    return loop.run_until_complete(coro)
  File "/home/docs/.pyenv/versions/3.7.3/lib/python3.7/asyncio/base_events.py", line 584, in run_until_complete
    return future.result()
  File "/home/docs/checkouts/readthedocs.org/user_builds/myst-nb/envs/latest/lib/python3.7/site-packages/nbclient/client.py", line 532, in async_execute
    self.nb.metadata['language_info'] = info_msg['content']['language_info']
KeyError: 'language_info'
```

its really weird because, because I can't replicate it locally, or even in [RTD PR Builds](https://docs.readthedocs.io/en/stable/guides/autobuild-docs-for-pull-requests.html), its literally only when it comes to build latest after a merge/commit 😕

davidbrochart:
Fixes #95

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -512,7 +512,13 @@ class NotebookClient(LoggingConfigurable):
             msg_id = await ensure_async(self.kc.kernel_info())
             info_msg = await self.async_wait_for_reply(msg_id)
             if info_msg is not None:
-                self.nb.metadata['language_info'] = info_msg['content']['language_info']
+                if 'language_info' in info_msg['content']:
+                    self.nb.metadata['language_info'] = info_msg['content']['language_info']
+                else:
+                    raise RuntimeError(
+                        'Kernel info received message content has no "language_info" key.'
+                        'Content is:\n' + str(info_msg['content'])
+                    )
             self.set_widgets_metadata()
 
         return self.nb


nbclient/client.py
```diff
@@ -512,7 +512,13 @@ async def async_execute(
             msg_id = await ensure_async(self.kc.kernel_info())
             info_msg = await self.async_wait_for_reply(msg_id)
             if info_msg is not None:
-                self.nb.metadata['language_info'] = info_msg['content']['language_info']
+                if 'language_info' in info_msg['content']:
+                    self.nb.metadata['language_info'] = info_msg['content']['language_info']
+                else:
+                    raise RuntimeError(
```
chrisjsewell, line 518: Should this raise an error or a warning, i.e. is `language_info` a strictly required component of the final notebook? In the specification https://nbformat.readthedocs.io/en/latest/format_description.html#notebook-metadata it is not present

chrisjsewell: I have limited knowledge of what this is actually doing lol,
but from:

https://github.com/chrisjsewell/jupyter_client/blob/778fa85c8ce7cd1a8ffe2fec2c1b8522d6c314ba/jupyter_client/adapter.py#L140
and
https://github.com/chrisjsewell/jupyter_client/blob/778fa85c8ce7cd1a8ffe2fec2c1b8522d6c314ba/jupyter_client/adapter.py#L251

it looks like if you are using the version 4 msg spec, it is not guaranteed, but if you use the version 5 msg spec it is (because of the `setdefault`)

davidbrochart: It seems to be [optional in the notebook](https://ipython.readthedocs.io/en/3.x/notebook/nbformat.html#top-level-structure), but [mandatory in the Jupyter protocol](https://jupyter-client.readthedocs.io/en/stable/messaging.html#kernel-info) from version 5.0:

*Changed in version 5.0: language_info, implementation, implementation_version, banner and help_links keys are added.*

So I guess we should handle it differently depending on the version? Good catch by the way!

chrisjsewell: no problem 😄

> So I guess we should handle it differently depending on the version? 

I guess so, although note the protocol did change many years ago I think, so this should be an edge case?

just baffled by why there is even an error on RTD, the only difference between these runs is that in the second I tried using your fork, which passes 🤷 (jupyter-client is the same version in both)

- https://readthedocs.org/projects/myst-nb/builds/11599522/
- https://readthedocs.org/projects/myst-nb/builds/11599497/

davidbrochart: Out of curiosity, can you try with nbclient master?

chrisjsewell: Yeh I tried and its the same for master and `0.4.1` tag.
I think I've spotted the difference though: https://github.com/readthedocs/readthedocs.org/issues/7360

MSeal: I don't see any reason to be backwards compatible to 4.x msg spec, so I think this is safe to make strict. 5 has been the defacto for a long time and I don't think there's many tools that work with 4 anymore.

MSeal:
Feel free to merge whenever -- we could add a test for this but I don't see it adding much benefit to capturing expected behavior in this error state.

Issue #99, opened 2020-08-22
Traitlets 5.x and async_start_new_kernel bug incompat


Carreau:
The following method has a bug:
```
    async def async_start_new_kernel(self, **kwargs) -> None:
        """Creates a new kernel.

        Parameters
        ----------
        kwargs :
            Any options for ``self.kernel_manager_class.start_kernel()``. Because
            that defaults to AsyncKernelManager, this will likely include options
            accepted by ``AsyncKernelManager.start_kernel()``, which includes ``cwd``.
        """
        assert self.km is not None
        resource_path = self.resources.get('metadata', {}).get('path') or None
        if resource_path and 'cwd' not in kwargs:
            kwargs["cwd"] = resource_path

        if hasattr(self.km, 'ipykernel') and self.km.ipykernel and self.ipython_hist_file:
            extra_arguments +=  ['--HistoryManager.hist_file={}'.format(self.ipython_hist_file)]

        await ensure_async(self.km.start_kernel(extra_arguments=extra_arguments, **kwargs))

```

It mutates `extra_arguments` which appends `HistoryManager.hist_file=...` each time it is called; this is incompatible with traitlets 5.0 wich will fail if multiple values have been passed for the same option.

MSeal:
It's backwards compatibility from nbconvert days as a convience function to control history that we hadn't ripped out of nbclient. Some nbconvert users had some patterns where they used it, though I don't know the justifications therein. Papermill sets ipython_hist_file so that this gets skipped if I recall correctly.

I'd be happy to move this to upstream applications if they care or remove in the future. For now I'm going to make a quick fix this evening and get a release shipped before the traitlets change ships.

MSeal:
0.4.2 release has the fix for this issue now

PR #100, opened 2020-08-25
Applied fix for duplicate HistoryManager setting being added

MSeal:
Fix for https://github.com/jupyter/nbclient/issues/99

I also realized when writing this that the reason we include it in nbclient is actually because ipykernel defaults to persisting all executions in sqlite unless to provide this argument, which can be a security concern in niche situations with a headless executor... plus it avoid sqlite permissions on systems where that occurs (mounted windows setups if I recall).


MSeal:
I'm going to merge this and get a patch release out since the change is pretty minor and it's a fix for the dependency coming out.

Carreau:
+1, thanks !

PR #105, opened 2020-09-11
Update README and Documentation

willingc:
Closes #84 

This PR updates:
- the README doc sections and prose
- fixes a broken link in CHANGELOG.md
- Updates sphinx docs to reflect the above changes

PR #106, opened 2020-09-11
Use rst not md link

willingc:
Should correct #105


willingc:
Since default role is set to any, we need to use  double backticks

```
# Set the default role so we can use `foo` instead of ``foo``
default_role = 'literal'
```

choldgraf:
thanks @willingc :-)

PR #107, opened 2020-09-11
Attempt to fix doc reference issue

MSeal:
Seeing if this fixes the reference lookup that started complaining


MSeal:
Closing because Carol beat me to it

PR #109, opened 2020-09-21
Cleanup references to preprocessor

takluyver:
There were a few places still using the ExecutePreprocessor names instead of the names in nbclient.

I also changed some examples to pass filenames to `nbformat.read()` and `nbformat.write()`. This is a shortcut that I'm pretty sure has worked for years.


davidbrochart (approved):

MSeal (approved):
Thanks!

PR #112, opened 2020-09-21
typo fix

akhmerov:
Formally execute preprocessor → Formerly execute preprocessor

Sorry, should have bundled this with the previous PR, but only spotted this now.


SylvainCorlay:
thx

PR #113, opened 2020-09-23
Experiment with threadpools

SylvainCorlay:
I am seeing many issues with nest-asyncio. Trying out threadpools.

### DIFF
--- a/nbclient/util.py
+++ b/nbclient/util.py
@@ -6,6 +6,8 @@
 import asyncio
 import sys
 import inspect
+import concurrent.futures
+
 from typing import Callable, Awaitable, Any, Union
 
 
@@ -28,6 +30,7 @@ def check_patch_tornado() -> None:
             tornado.concurrent.FUTURES = \
                 tornado.concurrent.FUTURES + (asyncio.Future, )  # type: ignore
 
+pool = concurrent.futures.ThreadPoolExecutor()
 
 def just_run(coro: Awaitable) -> Any:
     """Make the coroutine run, even if there is an event loop running (using nest_asyncio)"""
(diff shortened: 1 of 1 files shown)


nbclient/util.py
```diff
@@ -44,14 +47,9 @@ def just_run(coro: Awaitable) -> Any:
...
     if had_running_loop:
-        # if there is a running loop, we patch using nest_asyncio
-        # to have reentrant event loops
-        check_ipython()
-        import nest_asyncio
-        nest_asyncio.apply()
-        check_patch_tornado()
-    return loop.run_until_complete(coro)
-
+        return pool.submit(asyncio.run, coro).result()
```
davidbrochart, line 50: We could have only this line in the function, so that we always run in a thread with its own event loop.

nbclient/util.py
```diff
@@ -28,6 +30,7 @@ def check_patch_tornado() -> None:
             tornado.concurrent.FUTURES = \
                 tornado.concurrent.FUTURES + (asyncio.Future, )  # type: ignore
 
+pool = concurrent.futures.ThreadPoolExecutor()
```
davidbrochart, line 33: Any reason to use more than one thread? I think we should use `max_workers=1`.

SylvainCorlay: max_worker should be fine IMO.

davidbrochart: If you don't specify `mar_workers`, it will use [unnecessary resources](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor): "If max_workers is None or not given, it will default to the number of processors on the machine, multiplied by 5".
Since you never actually run concurrently, you don't need more than 1 thread.

Carreau:
closing as stale, we can reopen if necessary.

PR #117, opened 2020-10-10
Update kernel client class if it's the synchronous version

kevin-bates:
Users wishing to bring their own kernel manager implementations that also happen to include a kernel client implementation were unable to do so because NBClient was unconditionally adjusting the kernel client class to juptyer_client's `AsyncKernelClient` class.  This change makes that adjustment conditional only when the kernel manager's configured `kernel_client_class` specifies jupyter_client's older (and synchronous) `KernelClient`.  When that class is not currently configured on the kernel manager, whatever `kernel_client_class` that is configured will be used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# Pycharm stuff
+.idea/
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -341,7 +341,7 @@ class NotebookClient(LoggingConfigurable):
             self.km = self.kernel_manager_class(kernel_name=self.kernel_name, config=self.config)
 
         # If the current kernel manager is still using the default (synchronous) KernelClient class,
-        # switch to the async version since that's what NBClient expects.  Any custom kernel client 
+        # switch to the async version since that's what NBClient expects.  Any custom kernel client
         # classes must also support async method calls.
         if self.km.client_class == 'jupyter_client.client.KernelClient':
             self.km.client_class = 'jupyter_client.asynchronous.AsyncKernelClient'
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -341,8 +341,7 @@ class NotebookClient(LoggingConfigurable):
             self.km = self.kernel_manager_class(kernel_name=self.kernel_name, config=self.config)
 
         # If the current kernel manager is still using the default (synchronous) KernelClient class,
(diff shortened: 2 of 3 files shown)


kevin-bates:
I'm sorry, but I'm at a loss on how to address the `mypy` test failure.  I don't see other (closed) PRs with this issue but I suppose it could be a recent happening somehow.
```
tox: mypy
  mypy run-test-pre: PYTHONHASHSEED='0'
  mypy run-test: commands[0] | mypy nbclient/client.py nbclient/exceptions.py nbclient/__init__.py nbclient/jsonutil.py nbclient/output_widget.py nbclient/util.py nbclient/_version.py
  nbclient/util.py:26: error: Cannot find implementation or library stub for module named 'tornado.concurrent'
  nbclient/util.py:26: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
  nbclient/util.py:26: error: Cannot find implementation or library stub for module named 'tornado'
  Found 2 errors in 1 file (checked 7 source files)
  ERROR: InvocationError for command /home/runner/work/nbclient/nbclient/.tox/mypy/bin/mypy nbclient/client.py nbclient/exceptions.py nbclient/__init__.py nbclient/jsonutil.py nbclient/output_widget.py nbclient/util.py nbclient/_version.py (exited with code 1)
```

davidbrochart:
Yes, I think mypy checks more things in newer versions. Ca you try adding `# type: ignore` at the end of the line (with 2 spaces otherwise it is flake8 which won't be happy :smile:)?

davidbrochart:
> It is the user's responsibility to ensure any custom kernel client class implementations support the async methods that are available on `AsyncKernelClient`.

I'm not sure we should mention that (in the code comments as well), because in nbclient we run every kernel client's method using `ensure_async`, so they don't have to be async.

kevin-bates:
Thanks David, I've updated the comment(s) and addressed the `mypy` issue.

davidbrochart:
Thanks!

kevin-bates:
@davidbrochart @MSeal - would it be possible to get a patch release (0.5.1) or is one scheduled in the near future (like the end of next week)?

davidbrochart:
Sure, I'll make one tomorrow (if nobody beats me at it).

MSeal:
Done -- and I'm going to be offline starting tomorrow fyi :)

kevin-bates:
Thank you both!

davidbrochart:
I'd have expected that to be fast, but this is :rocket: fast, thanks @MSeal !

PR #122, opened 2020-11-05
Add `__reduce__` method to CellExecutionError

### LINKED ISSUE #121: CellExecutionError not unpickleable
## Issue
Unpickling a `CellExecutionError` exception raises 
```traceback
  File "<stdin>", line 1, in <module>
TypeError: __init__() missing 2 required positional
```
## Cause
https://github.com/jupyter/nbclient/blob/cac82a3fe6a6e1c2fff13d0b3a946536610a7405/nbclient/exceptions.py#L61-L65

As a subclass of `BaseException`, the `CellExecutionError` uses the default implementation of `__reduce__` that returns the values passed to `BaseException.__init__` followed by the instance dict of the exception.

E.g. the `__reduce__` result might look like
`(<class 'nbclient.exceptions.CellExecutionError'>, ('yikes',), {'traceback': 'yikes', 'ename': 'big', 'evalue': 'yikes'})`

Because of this, at unpickling time, only one argument is passed to the constructor (`"yikes"` in this example), whilst the attribute dictionary should be set later (after construction). Consequently, unpickling fails for `CellExecutionError` because it expects three positional arguments:

## Workaround:
For those who just want a solution, here's a function that can be used as a `concurrent.futures.Executor` initializer:
```python
def _patch_nbclient_exceptions():
    import nbclient.exceptions as exc

    # Fix https://github.com/jupyter/nbclient/issues/121
    def reduce_CellExecutionError(self):
        return type(self), (self.traceback, self.ename, self.evalue)

    exc.CellExecutionError.__reduce__ = reduce_CellExecutionError
```

## Solutions:
- Implement a custom `__reduce__` function that passes the object attributes as positional arguments only
- Implement default values for the other arguments (hack)
- Pass the attributes to the superclass init in the same order as they're defined in the `__init__` signature.

agoose77:
Fixes #121

### DIFF
--- a/nbclient/exceptions.py
+++ b/nbclient/exceptions.py
@@ -68,6 +68,9 @@ class CellExecutionError(CellControlSignal):
         self.ename = ename
         self.evalue = evalue
 
+    def __reduce__(self) -> tuple:
+        return type(self), (self.traceback, self.ename, self.evalue)
+
     def __str__(self) -> str:
         s = self.__unicode__()
         if not isinstance(s, str):


davidbrochart:
Thanks!

PR #130, opened 2021-02-07
Add allow_error_names option

### LINKED ISSUE #129: Allow specific errors to be ignored when executing cell
This is a feature request.

We are using Jupyter notebooks for [computational tutorials](https://github.com/NeuromatchAcademy/course-content) and have the somewhat odd need for our notebooks to execute while allowing a specific error class to be raised without halting.

The idea is that we ship notebooks with incomplete exercises that students are to fill out, and the exercise stubs raise a `NotImplementedError` to signal when something needs to be completed. We want to ignore those errors in the context of our CI workflow, where we are executing the notebooks using `nbconvert`.

I recently [solved this problem](https://github.com/NeuromatchAcademy/course-content/pull/542) by subclassing the `nbconvert.ExecutePreprocessor` class and overwriting a couple of relevant methods, which happen to be inherited from `nbclient`.

This was a little bit more complicated that it could have been, because we needed only to change the logic in `NotebookClient._check_raise_for_error`, but because that's a private method, I had to also copy the entire `async_execute_cell` method which calls it.

There are two ways that this could be easier for us:

- There is already a system for allowing errors in general. This could be extended to support our use-case of allowing a specific exception class.
- If the `_check_raise_for_error` method were public, it would simplify the process of writing a custom subclass that does bespoke error handling.

Thanks!

davidbrochart:
Fixes #129

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -101,10 +101,22 @@ class NotebookClient(LoggingConfigurable):
             """
             If ``False`` (default), when a cell raises an error the
             execution is stopped and a `CellExecutionError`
-            is raised.
+            is raised, except if the error name is in
+            ``allow_error_names``.
             If ``True``, execution errors are ignored and the execution
             is continued until the end of the notebook. Output from
             exceptions is included in the cell output in both cases.
+            """
+        ),
+    ).tag(config=True)
+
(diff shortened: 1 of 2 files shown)


MSeal (commented):
LGTM

PR #134, opened 2021-02-14
Don't depend on async_generator

### LINKED ISSUE #133: async-generator - why required in Python 3.9?
`Python 3.9.1`

Running `pipdeptree` I see:
```
* nbclient==0.5.2
 - async-generator [required: Any, installed: ?]
 ```

But according to https://github.com/python-trio/async_generator "This library gives you all that back to Python 3.5." which implies python 3.6+ already have async generator functionality.

Is it because `nbclient` is using `async-generator` on all versions of python3 for consistency?

davidbrochart:
Fixes #133

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -14,9 +14,6 @@ ignore_missing_imports = True
 [mypy-nest_asyncio.*]
 ignore_missing_imports = True
 
-[mypy-async_generator.*]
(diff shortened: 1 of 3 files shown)


davidbrochart:
We actually need `async-generator` for `asynccontextmanager`, which is only available from Python 3.7.

PR #137, opened 2021-02-25
Fix ipykernel's stop_on_error value

### LINKED ISSUE #136: Tests fail with the latest ipykernel
It seems that nbclient in a combination with the latest ipykernel 5.5.0 does not work correctly with `raises-exception` tag and test test_run_all_notebooks[Skip Exceptions with Cell Tags.ipynb-opts8] fails:
```
E           AssertionError: assert [{'name': 'stdout', 'output_type': 'stream', 'text': 'ok\n'}] == []
E             Left contains one more item: {'name': 'stdout', 'output_type': 'stream', 'text': 'ok\n'}
E             Full diff:
E             - []
E             + [{'name': 'stdout', 'output_type': 'stream', 'text': 'ok\n'}]
```
It works well for me with ipykernel 5.4.3 so I've reported it to ipykernel: https://github.com/ipython/ipykernel/issues/598
But the truth is that I'm not able to reproduce the same behavior in jupyter notebook so it might not be an error in ipykernel.

davidbrochart:
Fixes #136
Closes https://github.com/ipython/ipykernel/issues/598

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -807,11 +807,16 @@ class NotebookClient(LoggingConfigurable):
             cell['metadata']['execution'] = {}
 
         self.log.debug("Executing cell:\n%s", cell.source)
+
+        cell_allows_errors = (not self.force_raise_errors) and (
+            self.allow_errors
+            or "raises-exception" in cell.metadata.get("tags", []))
+
         parent_msg_id = await ensure_async(
             self.kc.execute(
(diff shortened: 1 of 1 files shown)


MSeal:
Ty for tracking that down! We should do a patch release this weekend for this

davidbrochart:
I released nbclient v0.5.3.

PR #139, opened 2021-03-07
windows py38 and py39

### LINKED ISSUE #85: Asyncio Python 3.8
See this related issue: https://github.com/nteract/papermill/issues/515, but essentially the async implementation on windows 3.8 uses a different implementation that doesn't include all the same functions. This means pyzqm barfs on loop actions. We likely need to patch nbclient in a similar way? I know this should only really be set on entrypoint but from testing setting it multiple times works just fine and the reality is there is no other event loop policy right now that we'd be worried about overwriting.

### LINKED ISSUE #125: Re-enable CI for Windows with Python 3.8 and 3.9
To get CI to pass Windows has been excluded from GitHub Actions testing. If root cause of Windows failure is identified, then CI can be included again.

juhuebner:
fixes #85, fixes #125

* patched according to https://github.com/tornadoweb/tornado/issues/2608#issuecomment-491489432
  * added version selection and type ignore 
* included windows python 3.8 and python 3.9
* updated tox for multiple platform usage
* added type ignore for version_info < (3, 7)
  * the type ignore should be made os dependent

forgot to squash my dev branch ;)

poss. 
* fixes https://github.com/executablebooks/jupyter-book/issues/958
* fixes https://github.com/executablebooks/jupyter-book/issues/906

### DIFF
--- a/nbclient/__init__.py
+++ b/nbclient/__init__.py
@@ -13,5 +13,5 @@ def _cleanup() -> None:
 # the fix for python3.7: https://github.com/python/cpython/pull/15706/files
 if sys.platform == 'win32':
     if sys.version_info < (3, 7):
-        subprocess._cleanup = _cleanup
-        subprocess._active = None
+        subprocess._cleanup = _cleanup  # type: ignore
+        subprocess._active = None  # type:ignore
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -33,7 +33,7 @@ from .util import run_sync, ensure_async
 from .output_widget import OutputWidget
 
 
-if os_name = "nt":
+if os_name == "nt":
     # patch c.f. https://github.com/tornadoweb/tornado/issues/2608#issuecomment-491489432
     asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())  # python-3.8.x
 
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -1,3 +1,4 @@
+from os import name as os_name
 import atexit
 import collections
 import datetime
@@ -32,8 +33,9 @@ from .util import run_sync, ensure_async
 from .output_widget import OutputWidget
 
 
-# patch c.f. https://github.com/tornadoweb/tornado/issues/2608#issuecomment-491489432
-asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())  # python-3.8.x
+if os_name = "nt":
+    # patch c.f. https://github.com/tornadoweb/tornado/issues/2608#issuecomment-491489432
+    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())  # python-3.8.x
 
 
 def timestamp() -> str:
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -32,6 +32,10 @@ from .util import run_sync, ensure_async
 from .output_widget import OutputWidget
 
 
+# patch c.f. https://github.com/tornadoweb/tornado/issues/2608#issuecomment-491489432
+asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())  # python-3.8.x
+
+
 def timestamp() -> str:
     return datetime.datetime.utcnow().isoformat() + 'Z'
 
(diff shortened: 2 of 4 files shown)


tox.ini
```diff
@@ -39,13 +41,22 @@ commands =
...
-[testenv:dist]
+[testenv:dist-{win, linux}]
 skip_install = true
-# Have to use /bin/bash or the `*` will cause that argument to get quoted by the tox command line...
+platform =
+    win: win|msys
+    linux: linux|darwin
+allowlist_externals = 
+    win: powershell
+    linux: bash
```
davidbrochart, line 51: Could you explain why we need this?

tox.ini
```diff
@@ -39,13 +41,22 @@ commands =
...
+
 commands =
     python setup.py sdist --dist-dir={distdir} bdist_wheel --dist-dir={distdir}
-    /bin/bash -c 'python -m pip install -U --force-reinstall {distdir}/nbclient*.whl'
-    /bin/bash -c 'python -m pip install -U --force-reinstall --no-deps {distdir}/nbclient*.tar.gz'
+    win: powershell Get-ChildItem -Path {distdir} -Filter nbclient*.whl | foreach \{python -m pip install -U --force-reinstall $_.FullName\}
+    win: powershell Get-ChildItem -Path {distdir} -Filter nbclient*.tar.gz | foreach \{python -m pip install -U --force-reinstall --no-deps $_.FullName\}
+    # Have to use /bin/bash or the `*` will cause that argument to get quoted by the tox command line...
+    linux: bash -c 'python -m pip install -U --force-reinstall {distdir}/nbclient*.whl'
+    linux: bash -c 'python -m pip install -U --force-reinstall --no-deps {distdir}/nbclient*.tar.gz'
```
davidbrochart, line 59: We don't need to use `/bin/bash` anymore?

davidbrochart:
Thanks a lot @juhuebner, I made some comments mainly for me to understand better your changes.

tox.ini
```diff
@@ -1,13 +1,15 @@
 [tox]
 skipsdist = true
-envlist = py{36,37,38, 39}, flake8, mypy, dist, manifest, docs
+envlist = py{36,37,38, 39}, flake8, mypy, dist-{win, linux}, manifest, docs
+# requires =
+#     tox-factor
```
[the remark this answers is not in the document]
juhuebner: > Left-over?

Yes, this can go
```
# requires =
#     tox-factor
```

tox.ini
juhuebner: The bash should be in the path. Even cleaner would be to use just `sh` 
`linux: sh -c 'python -m pip install -U --force-reinstall {distdir}/nbclient*.whl'` etc.

tox.ini
juhuebner: The wildcard usage like `nbclient*.whl` is quite difficult to get into windows, even the gh windows runners have the bash installed. It's all about what is replaced within the quotes. So, what I did here was to distangle Windows and POSIX like systems. `tox `will check the platform via `sys.platform` and select the correct environment. Via that it also runs, e.g., on Windows 10 systems, where there is not a bash installed in general.

It might be a bit blown up to get this one `whl` file into `pip install` but I haven't found another way.

nbclient/__init__.py
```diff
@@ -13,5 +13,5 @@ def _cleanup() -> None:
 # the fix for python3.7: https://github.com/python/cpython/pull/15706/files
 if sys.platform == 'win32':
     if sys.version_info < (3, 7):
-        subprocess._cleanup = _cleanup
-        subprocess._active = None
+        subprocess._cleanup = _cleanup  # type: ignore
+        subprocess._active = None  # type:ignore
```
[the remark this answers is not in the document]
juhuebner: mypy will raise an error for py39 since these methods aren't implemented anymore.

nbclient/client.py
```diff
@@ -32,6 +34,11 @@
 from .output_widget import OutputWidget
 
 
+if os_name == "nt" and version_info.major == 3 and version_info.minor >= 8:
```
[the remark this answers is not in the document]
juhuebner: changed `if os_name == "nt" and version_info.major >= (3, 8):` to `if os_name == "nt" and version_info >= (3, 8):` in new commit.

Carreau:
closing as stale

PR #146, opened 2021-04-27
Test ipykernel master

davidbrochart:


davidbrochart:
The failures are due to ipython>=7.21.0 not being installable in python 3.6, and other errors in ipykernel master that should be fixed before the release. Apart from that, setting `IPYKERNEL_CELL_NAME=<IPY-INPUT>` fixes our tests with strict output check.

PR #151, opened 2021-08-04
Add `skip-execution` cell tag functionality

chrisjsewell:
This is something that has been requested and up-voted in https://github.com/executablebooks/jupyter-book/issues/833

I guess this is the simplest implementation, although happy to discuss (a) if you would back this idea in the first place and (b) is you think any more configuration is needed (e.g. at the traitlets level)

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -802,6 +802,10 @@ class NotebookClient(LoggingConfigurable):
         if cell.cell_type != 'code' or not cell.source.strip():
             self.log.debug("Skipping non-executing cell %s", cell_index)
             return cell
+        
+        if "skip-execution" in cell.metadata.get("tags", []):
+            self.log.debug("Skipping tagged cell %s", cell_index)
+            return cell
 
         if self.record_timing and 'execution' not in cell['metadata']:
             cell['metadata']['execution'] = {}
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -812,7 +812,7 @@ class NotebookClient(LoggingConfigurable):
         if cell.cell_type != 'code' or not cell.source.strip():
             self.log.debug("Skipping non-executing cell %s", cell_index)
             return cell
-        
+
         if self.skip_cells_with_tag in cell.metadata.get("tags", []):
             self.log.debug("Skipping tagged cell %s", cell_index)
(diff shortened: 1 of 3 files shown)


chrisjsewell:
I guess this is also related to https://discourse.jupyter.org/t/is-there-a-cell-tag-convention-to-skip-execution/5445, but nothing really come of that, so I wanted to be a bit more direct 😬 

I guess you could add a traitlets options, which disabled this feature by default; so it is not a potentially breaking change for anyone else, but allowed us folks that want do want to have such functionality to use it.

davidbrochart:
I think it can be useful, but maybe we could make it more generic? For instance, if we decide to make it configurable through a trait, we could have one such as `skip_cells_with_tag=skip-execution`, what do you think?

chrisjsewell:
>  we could have one such as skip_cells_with_tag=skip-execution, what do you think?

that works for me cheers 😄 will update the PR in a bit...

chrisjsewell:
Hey @davidbrochart I have updated the PR with the `skip_cells_with_tag` trait

chrisjsewell:
oops fixed flake8 (fancy me adding a pre-commit config for that and maybe mypy?)

davidbrochart:
> fancy me adding a pre-commit config for that and maybe mypy?

Sure, thanks.

chrisjsewell:
> Sure, thanks.

I'll do that in another PR ta

nbclient/client.py
```diff
@@ -139,6 +139,16 @@ class NotebookClient(LoggingConfigurable):
         ),
     ).tag(config=True)
 
+    skip_cells_with_tag: str = Unicode(
+        'skip-execution',
+        help=dedent(
+            """
+            Name of the cell tag to use to,
+            to denote a cell that should be skipped.
```
davidbrochart, line 147:
```suggestion
            Name of the cell tag to use to denote a cell that should be skipped.
```

davidbrochart:
@takluyver we are seeing the same error as in #159 (`zmq.error.ZMQError: Address already in use`).

davidbrochart (approved):

davidbrochart:
Awesome, thanks!

PR #152, opened 2021-08-09
Replace km.cleanup with km.cleanup_resources

davidbrochart:
`cleanup` was removed in jupyter_client=7.0


blink1073 (approved):
Thank you!

PR #160, opened 2021-09-15
PoP for Inline variable insertion in markdown

chrisjsewell:
This is a hacked together Proof-of-Principle for https://discourse.jupyter.org/t/inline-variable-insertion-in-markdown/10525/87

nbclient would accept an additional configuration trait: `parse_md_expressions`, which is a callable that takes a Markdown string and extracts from it a list of substitution expressions.

If present, these are then parsed to the kernel, via `user_expressions`, to return their mime-bundles, which are then stored as attachments on the Markdown cell

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -32,6 +32,8 @@ def timestamp() -> str:
     return datetime.datetime.utcnow().isoformat() + 'Z'
 
 
+MD_EXPRESSIONS_PREFIX = "md-expr"
+
 class NotebookClient(LoggingConfigurable):
     """
     Encompasses a Client for executing cells in a notebook
@@ -1030,7 +1032,7 @@ class NotebookClient(LoggingConfigurable):
             self.kc.execute(
                 '',
                 silent=True,
-                user_expressions={f"md-expr-{i}": expr for i, expr in enumerate(expressions)},
+                user_expressions={f"{MD_EXPRESSIONS_PREFIX}-{i}": expr for i, expr in enumerate(expressions)},
             )
         )
         task_poll_kernel_alive = asyncio.ensure_future(
@@ -1061,7 +1063,10 @@ class NotebookClient(LoggingConfigurable):
                 raise
         self._check_raise_for_error(cell, exec_reply)
         attachments = {key: val["data"] for key, val in exec_reply["content"]["user_expressions"].items()}
-        cell.setdefault("attachments", {}).update(attachments)
+        cell.setdefault("attachments", {})
+        # remove old expressions from cell
+        cell["attachments"] = {key: val for key, val in cell["attachments"].items() if not key.startswith(MD_EXPRESSIONS_PREFIX)}
+        cell["attachments"].update(attachments)
         self.nb['cells'][cell_index] = cell
         return cell
 
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -14,7 +14,18 @@ from jupyter_client import KernelManager
 from jupyter_client.client import KernelClient
 from nbformat import NotebookNode
 from nbformat.v4 import output_from_msg
-from traitlets import Any, Bool, Callable, Dict, Enum, Integer, List, Type, Unicode, default
+from traitlets import (
+    Any,
+    Bool,
+    Callable,
+    Dict,
+    Enum,
+    Integer,
+    List,
+    Type,
+    Unicode,
+    default,
+)
 from traitlets.config.configurable import LoggingConfigurable
 
 from .exceptions import (
@@ -33,6 +44,7 @@ def timestamp() -> str:
 
 
 MD_EXPRESSIONS_PREFIX = "md-expr"
+
 
 class NotebookClient(LoggingConfigurable):
     """
@@ -294,7 +306,7 @@ class NotebookClient(LoggingConfigurable):
             """
             A function to extract expression variables from a Markdown cell.
             """
-        )
+        ),
     )
 
     resources: t.Dict = Dict(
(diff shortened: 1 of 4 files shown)


chrisjsewell:
Oh wait a minute, its obviously in the exec_reply 🤦
thats ok I will carry on messing around with this

chrisjsewell:
working example done 😄 
Obviously very rough around the edges, but feel free to give feedback already (I'll post about it in the discussion now)

meeseeksmachine:

https://discourse.jupyter.org/t/inline-variable-insertion-in-markdown/10525/90

nbclient/client.py
```diff
@@ -1033,6 +1051,104 @@ def handle_comm_msg(
...
+        except Exception as e:
+            # Best effort to cancel request if it hasn't been resolved
+            try:
+                # Check if the task_poll_output is doing the raising for us
+                if not isinstance(e, CellControlSignal):
+                    task_poll_expr_msg.cancel()
+            finally:
+                raise
+        self._check_raise_for_error(cell, exec_reply)
+        attachments = {key: val["data"] if "data" in val else {"traceback": "\n".join(val["traceback"])} for key, val in exec_reply["content"]["user_expressions"].items()}
```
agoose77, line 1089: Over on `jupyterlab-imarkdown` [I'm playing around with](https://github.com/agoose77/jupyterlab-imarkdown/blob/ed6d3dfe6b6edd5947ee824dbf131bae6fe70a1b/src/kernel.ts#L51-L65 ) keeping the kernel-output itself as [two different MIME types](https://github.com/agoose77/jupyterlab-imarkdown/blob/ed6d3dfe6b6edd5947ee824dbf131bae6fe70a1b/src/attachment.ts#L3-L4), i.e. the "attachment" is just one of two possible MIME types.

My thought process here is that we keep the structural information, and do the formatting [as late as possible](https://github.com/agoose77/jupyterlab-imarkdown/blob/master/src/cell.ts#L60-L83). The `status` field probably shouldn't be stored, as it's already encoded in the MIME type.

How do you feel about doing something like this?

chrisjsewell: Heya 👋🏼 well my feeling is that really we use have a new cell type with a new json schema, that would be a sort of hybrid between code and markdown; e.g. it should definitely have an execution number and then this data should have its own “expressions” key, where yeh we would have a more 1-to-1 mapping with the data returned by the kernel

chrisjsewell: Having these special mime types indeed works, but I feel is a bit “hacky” for core jupyter

agoose77:
Hey @chrisjsewell, thanks for linking me to this!

In general, I'm glad that you're tackling the other "big" Jupyter frontend that needs to implement this functionality. 

One of the problems as I see it with the existing notebook schema is that [we only support GFM](https://jupyter-notebook.readthedocs.io/en/4.x/examples/Notebook/rstversions/Working%20With%20Markdown%20Cells.html#github-flavored-markdown-gfm). Because we are deviating from that to implement the concept of "inline-variable" insertion, we are defining our own "flavour". For this reason, I think it's a good idea that you've used a configurable to off-load the implementation. In the long run, I think we need to standardise this somehow. [One of my ideas](https://discourse.jupyter.org/t/notebook-cell-type-generaliastion/10703) is to remove Markdown as a special-case from the Jupyter Notebook schema, and instead have a MIME-type associated with each non-cell. This could then be used to provide finer-grained information about the Markdown flavour(s) used. This is not really in-scope for this particular PoC, though, so I won't keep musing on it in this thread!

nbclient/client.py
chrisjsewell: so it would be along the lines of:

```json
  {
   "cell_type": "imarkdown",
   "execution_count": 1,
   "metadata": {},
   "expressions": [
      {
        "output_type": "display",
        "data": {
          "text/plain": "1"
        }
      },
      {
        "output_type": "error",
        "traceback": ["..."]
      },
   ],
   "source": [
    "{{ a }}\n"
   ]
  }
```

meeseeksmachine:

https://discourse.jupyter.org/t/inline-variable-insertion-in-markdown/10525/113

davidbrochart:
@chrisjsewell are you still planning to work on this PR?

davidbrochart:
Thanks @chrisjsewell, I like the idea of using `user_expressions`.
One thing I don't understand though, would it be a code cell, or a markdown cell?

chrisjsewell:
> One thing I don't understand though, would it be a code cell, or a markdown cell?

On a code-cell.
Essentially, forget the initial proposal regarding any interaction with the markdown cell; at least for now that can be done in pre/post-processing steps on the notebook (outside nbclient)

Basically what we would be offering to nbclient users, is the ability to ask the kernel to snapshot a set of variable outputs, after execution of the code source code, without having to display them (and in a kernel agnostic manner)

choldgraf:
I was looking through the `user_expressions` docs (of which there are very little 🤦) and I think that supporting `user_expressions` in `nbclient` in this way is a nice step forward here. It follows the `jupyter_client` spec nicely, and I sorta think about `nbclient` like `jupyter_client` extended to notebooks. An added bonus here is that it gives us a foundation that we could use to play around with the markdown insertion stuff. So I'm +1 supporting `user_expressions` in `nbclient`.

davidbrochart:
> that can be done in pre/post-processing steps on the notebook (outside nbclient)

@chrisjsewell you might want to have a look at #188, which adds `on_cell_start` and `on_cell_complete` hooks. Could they be used for the pre/post processing steps you're talking about?

chrisjsewell:
FYI, I went ahead and made a PoP PR in myst-nb for full inline variable evaluation 😄 https://myst-nb--382.org.readthedocs.build/en/382/use/inline_execution.html
Basically, its running a `NotebookClient`, as it parses the document (within a context manager); calling kernel execution whenever it meets a code cell or inline variable.

chrisjsewell:
Closing this, I actually went a different route: https://myst-nb.readthedocs.io/en/latest/render/inline.html

May be good to eventually upstream the additional code in:  https://github.com/executablebooks/MyST-NB/blob/master/myst_nb/ext/eval/__init__.py

PR #164, opened 2021-09-26
Added Pipfile stuff to gitignore

palewire:
If this project is all in on requirements.txt, we should ignore any Pipfiles. This will allow devs who prefer to use that approach to work without accidentally littering their files in the repo.


MSeal (approved):

MSeal:
ty!

PR #166, opened 2021-10-01
Fix tests for ipykernel without debugpy

frenzymadness:
It is supported by ipykernel to run it without debugpy installed
but it produces warnings informing about disabled debugging.
Those warnings should not cause tests of nbclient to fail.

See: https://github.com/ipython/ipykernel/blob/1ba6b48a97877ff7a564af32c531618efb7d2a57/ipykernel/ipkernel.py#L36

### DIFF
--- a/nbclient/tests/test_client.py
+++ b/nbclient/tests/test_client.py
@@ -239,6 +239,19 @@ def notebook_resources():
     return {'metadata': {'path': os.path.join(current_dir, 'files')}}
 
 
+def filter_messages_on_error_output(err_output):
+    filtered_result = []
+    allowed_lines = [
+        # ipykernel migh be installed without debugpy extension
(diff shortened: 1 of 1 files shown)


davidbrochart (approved):

SylvainCorlay:
I am fine with this change, although still a bit unsure of how much the code base should take into account that some package managers remove a dependency.

frenzymadness:
The pre-commit check complains about a file I haven't touched here. Also, the failure of 3.9 on macos seems unrelated.


> I am fine with this change, although still a bit unsure of how much the code base should take into account that some package managers remove a dependency.

I understand. In Fedora, we are running tests in RPM build and because we ship ipykernel without debugpy, the tests are failing. We can either fix it here or on the RPM level and because the fix is easy and might be useful also for other warnings/errors in the future, I've proposed it.

PR #169, opened 2021-10-29
fix: updating buffers overwrote previous buffers

maartenbreddels:
We now add the buffers, and overwrite old ones based on the path.

For instance, if you did in the notebook:
```
mark.x = np.arange(10)
mark.y = np.arange(1, 11)  # this would clear the buffer for x
```

Now we keep the buffer for x and y.

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -334,7 +334,7 @@ class NotebookClient(LoggingConfigurable):
         self.code_cells_executed = 0
         self._display_id_map = {}
         self.widget_state: t.Dict[str, t.Dict] = {}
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Thanks for helping out!

davidbrochart:
It think the CI failures are not related (it starting to be annoying, we fix them in another PR).

choldgraf (approved):
I don't understand the widget architecture well enough to really know if this is the right implementation, but I trust y'alls knowledge there, and see a nice test so I am +1 on this :-)

davidbrochart:
Thanks @maartenbreddels !

PR #172, opened 2021-11-09
Move IPYKERNEL_CELL_NAME from tox to pytest

frenzymadness:
This allows pytest to be executed directly without tox and without
manual setting of the variable. Tox works well as before.

Possible fix for #157


davidbrochart:
Thanks!

PR #188, opened 2021-12-28
Client hooks

devintang3:
Continuation of https://github.com/jupyter/nbclient/pull/79

I rebased with master and added some tests. I occasionally get errors on `test_many_parallel_notebooks` which I don't think is related, but I'll also get it on `test_error_cell_hooks` in py38. It's an intermittent error that gets resolved if I add a sleep, so I think it's something related to the async task not running right away. Any guidance here would be appreciated.

I also removed py36 from tox since it looked like the minimum version is now Python3.7 and it was just hanging when I tried running it.

### DIFF
--- a/docs/client.rst
+++ b/docs/client.rst
@@ -96,6 +96,36 @@ on both versions. Here the traitlet ``kernel_name`` helps simplify and
 maintain consistency: we can just run a notebook twice, specifying first
 "python2" and then "python3" as the kernel name.
 
+Hooks before and after notebook or cell execution
+-------------------------------------------------
+There are several configurable hooks that allow the user to execute code before and
+after a notebook or a cell is executed. Each one is configured with a function that will be called in its
+respective place in the execution pipeline.
+Each is described below:
+
+**Notebook-level hooks**: These hooks are called with a single extra parameter:
+
+- ``notebook=NotebookNode``: the current notebook being executed.
+
+Here is the available hooks:
+
+- ``on_notebook_start`` will run when the notebook client is initialized, before any execution has happened.
+- ``on_notebook_complete`` will run when the notebook client has finished executing, after kernel cleanup.
+- ``on_notebook_error`` will run when the notebook client has encountered an exception before kernel cleanup.
+
+**Cell-level hooks**: These hooks are called with two parameters:
+
+- ``cell=NotebookNode``: a reference to the current cell.
+- ``cell_index=int``: the index of the cell in the current notebook's list of cells.
+
+Here are the available hooks:
+
+- ``on_cell_start`` will run for all cell types before the cell is executed.
+- ``on_cell_execute`` will run right before the code cell is executed.
+- ``on_cell_complete`` will run after execution, if the cell is executed with no errors.
+- ``on_cell_error`` will run if there is an error during cell execution.
+
+
 Handling errors and exceptions
 ------------------------------
 
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -26,7 +26,7 @@ from .exceptions import (
     DeadKernelError,
 )
 from .output_widget import OutputWidget
-from .util import ensure_async, run_sync
+from .util import ensure_async, run_sync, run_hook
 
 
 def timestamp(msg: Optional[Dict] = None) -> str:
@@ -261,6 +261,45 @@ class NotebookClient(LoggingConfigurable):
 
     kernel_manager_class: KernelManager = Type(config=True, help='The kernel manager class to use.')
 
+    on_execution_start: t.Optional[t.Callable] = Any(
+        default_value=None,
+        allow_none=True,
+        help=dedent("""
+        Called after the kernel manager and kernel client are setup, and cells
+        are about to execute.
+        Called with kwargs `kernel_id`.
+        """),
+    ).tag(config=True)
+
+    on_cell_start: t.Optional[t.Callable] = Any(
+        default_value=None,
+        allow_none=True,
+        help=dedent("""
+        A callable which executes before a cell is executed.
+        Called with kwargs `cell`, and `cell_index`.
+        """),
+    ).tag(config=True)
+
+    on_cell_complete: t.Optional[t.Callable] = Any(
+        default_value=None,
+        allow_none=True,
+        help=dedent("""
+        A callable which executes after a cell execution is complete. It is
+        called even when a cell results in a failure.
+        Called with kwargs `cell`, and `cell_index`.
+        """),
+    ).tag(config=True)
+
+    on_cell_error: t.Optional[t.Callable] = Any(
+        default_value=None,
+        allow_none=True,
+        help=dedent("""
+        A callable which executes when a cell execution results in an error.
+        This is executed even if errors are suppressed with `cell_allows_errors`.
+        Called with kwargs `cell`, and `cell_index`.
+        """),
+    ).tag(config=True)
+
     @default('kernel_manager_class')
     def _kernel_manager_class_default(self) -> KernelManager:
         """Use a dynamic default to avoid importing jupyter_client at startup"""
@@ -442,6 +481,7 @@ class NotebookClient(LoggingConfigurable):
             await self._async_cleanup_kernel()
             raise
         self.kc.allow_stdin = False
+        run_hook(sself.on_execution_start)
         return self.kc
 
     start_new_kernel_client = run_sync(async_start_new_kernel_client)
@@ -745,7 +785,11 @@ class NotebookClient(LoggingConfigurable):
             return True
         return False
 
-    def _check_raise_for_error(self, cell: NotebookNode, exec_reply: t.Optional[t.Dict]) -> None:
+    def _check_raise_for_error(
+            self,
+            cell: NotebookNode,
+            cell_index: int,
+            exec_reply: t.Optional[t.Dict]) -> None:
 
         if exec_reply is None:
             return None
@@ -760,8 +804,10 @@ class NotebookClient(LoggingConfigurable):
             or "raises-exception" in cell.metadata.get("tags", [])
         )
 
-        if not cell_allows_errors:
-            raise CellExecutionError.from_cell_and_msg(cell, exec_reply_content)
+        if (exec_reply is not None) and exec_reply['content']['status'] == 'error':
+            run_hook(self.on_cell_error, cell=cell, cell_index=cell_index)
+            if self.force_raise_errors or not cell_allows_errors:
+                raise CellExecutionError.from_cell_and_msg(cell, exec_reply['content'])
 
     async def async_execute_cell(
         self,
@@ -821,11 +867,13 @@ class NotebookClient(LoggingConfigurable):
             self.allow_errors or "raises-exception" in cell.metadata.get("tags", [])
         )
 
+        run_hook(self.on_cell_start, cell=cell, cell_index=cell_index)
         parent_msg_id = await ensure_async(
             self.kc.execute(
                 cell.source, store_history=store_history, stop_on_error=not cell_allows_errors
             )
         )
+        run_hook(self.on_cell_complete, cell=cell, cell_index=cell_index)
         # We launched a code cell to execute
         self.code_cells_executed += 1
         exec_timeout = self._get_timeout(cell)
@@ -859,7 +907,7 @@ class NotebookClient(LoggingConfigurable):
 
         if execution_count:
             cell['execution_count'] = execution_count
-        self._check_raise_for_error(cell, exec_reply)
+        self._check_raise_for_error(cell, cell_index, exec_reply)
         self.nb['cells'][cell_index] = cell
         return cell
 
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -15,7 +15,18 @@ from jupyter_client import KernelManager
 from jupyter_client.client import KernelClient
 from nbformat import NotebookNode
 from nbformat.v4 import output_from_msg
-from traitlets import Any, Bool, Dict, Enum, Integer, List, Type, Unicode, default
+from traitlets import (
+    Any,
+    Bool,
+    Callable,
+    Dict,
+    Enum,
+    Integer,
+    List,
+    Type,
+    Unicode,
+    default,
+)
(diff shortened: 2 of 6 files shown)


nbclient/util.py
```diff
@@ -102,3 +103,15 @@ async def ensure_async(obj: Union[Awaitable, Any]) -> Any:
...
+def run_hook(hook: Optional[Callable], **kwargs) -> None:
+    if hook is None:
+        return
+    if inspect.iscoroutinefunction(hook):
+        future = hook(**kwargs)
+    else:
+        loop = asyncio.get_event_loop()
+        hook_with_kwargs = partial(hook, **kwargs)
+        future = loop.run_in_executor(None, hook_with_kwargs)
+    asyncio.ensure_future(future)
```
davidbrochart, line 117: I don't understand, does it mean that hooks are going to be launched in the background? BTW, it would be nice to test async hooks.
I think `run_hook` should be `async`:
```python
async def run_hook(hook: Optional[Callable], **kwargs) -> None:
    await ensure_async(hook(**kwargs))
```

davidbrochart (changes requested):
Thanks a lot for this PR @devintang3.
Would you like to add these new features to the documentation also?

nbclient/util.py
devintang3: Yeah, the intent is for the hook to be launched in the background.

devintang3:
Thanks for the suggestions. I've applied your recommendations as well as added more tests. 
Some changes to note:
* tests/util.py has been renamed to tests/test_util.py so that pytest can pick it up automatically
* Added the package pytest-asyncio to requirements-dev.txt so I can test some async methods
* Updated docs
* The pre-commit hook was complaining about unused variables so I've removed them from test_client.py. I'm not quite sure what the purpose of the `loop = asyncio.get_event_loop()` was

nbclient/client.py
```diff
@@ -245,6 +245,52 @@ class NotebookClient(LoggingConfigurable):
...
     kernel_manager_class: KernelManager = Type(config=True, help='The kernel manager class to use.')
 
+    on_execution_start: t.Optional[t.Callable] = Any(
+        default_value=None,
+        allow_none=True,
+        help=dedent(
+            """
+        Called after the kernel manager and kernel client are setup, and cells
+        are about to execute.
+        """
```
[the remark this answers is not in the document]
davidbrochart: You didn't make the change.

nbclient/util.py
```diff
@@ -102,3 +103,15 @@ async def ensure_async(obj: Union[Awaitable, Any]) -> Any:
...
+async def run_hook(hook: Optional[Callable], **kwargs) -> None:
+    if hook is None:
+        return
+    if asyncio.iscoroutinefunction(hook):
+        future = hook(**kwargs)
+    else:
+        loop = asyncio.get_event_loop()
+        hook_with_kwargs = partial(hook, **kwargs)
+        future = loop.run_in_executor(None, hook_with_kwargs)
+    await asyncio.ensure_future(future)
```
davidbrochart, line 117:
```suggestion
    res = hook(**kwargs)
    if inspect.isawaitable(res):
        await res
```

davidbrochart (changes requested):

davidbrochart:
Could you rebase on master? It looks like you merged master into your branch instead.

nbclient/client.py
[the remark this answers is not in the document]
devintang3: Thanks! Not sure how I missed this one.

nbclient/client.py
```diff
@@ -805,11 +865,13 @@ async def async_execute_cell(
             self.allow_errors or "raises-exception" in cell.metadata.get("tags", [])
         )
 
+        await run_hook(self.on_cell_start, cell=cell, cell_index=cell_index)
```
chrisjsewell, line 868: while we are at it, perhaps there could be a hook at the very start of this function, before even skipping non-executing cells, like `on_cell_visit`

nbclient/client.py
```diff
@@ -426,6 +483,7 @@ async def async_start_new_kernel_client(self) -> KernelClient:
             await self._async_cleanup_kernel()
             raise
         self.kc.allow_stdin = False
+        await run_hook(self.on_execution_start)
```
choldgraf, line 486:
```suggestion
        await run_hook(self.on_execution_start, notebook=self.nd)
```

What about something like this so that people have access to the NotebookNode before execution?

nbclient/client.py
```diff
@@ -245,6 +256,52 @@ class NotebookClient(LoggingConfigurable):
 
     kernel_manager_class: KernelManager = Type(config=True, help='The kernel manager class to use.')
 
+    on_execution_start: t.Optional[t.Callable] = Callable(
```
choldgraf, line 259: I think `on_notebook_start` might be more intuitive. `on_execution_start` makes me think that it will be called as soon as the execution begins, whereas this is actually called when the notebook is initialized but no cells have been identified for execution yet.

nbclient/client.py
choldgraf, line 486: For symmetry's sake, why not also include two extra notebook-level hooks: `on_notebook_complete` and `on_notebook_error`, that map on to the two cell execution hooks but at a notebook level?

That way you would have the same basic pattern for both the notebook, and for each cell in the notebook:

- hook for just before execution happens (`on_notebook_start`, `on_cell_start`)
- hook for when execution completes (`on_notebook_complete`, `on_cell_complete`)
- hook for when execution has an error (`on_notebook_error`, `on_cell_error`)

choldgraf (commented):
I think this is a really nice extension for `nbclient` - I have a few comments and suggestions, mostly around the API design and the documentation to make it clear how to use it

nbclient/client.py
devintang3: Does it make sense to just move `on_cell_start` to the start instead of creating a new hook?

docs/client.rst
```diff
@@ -96,6 +96,15 @@ on both versions. Here the traitlet ``kernel_name`` helps simplify and
...
 
+In addition to the two above, we also support traitlets for hooks. They are as
+follows: ``on_execution_start``, ``on_cell_start``, ``on_cell_complete``,
+``on_cell_error``. These traitlets allow specifying a ``Callable`` function,
+which will run at certain points during the notebook execution and is executed asynchronously.
+``on_execution_start`` will run when the notebook client is kicked off.
+``on_cell_start`` will run right before each cell is executed.
+``on_cell_complete`` will run right after the cell is executed.
+``on_cell_error`` will run if there is an error in the cell.
+
```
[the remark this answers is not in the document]
davidbrochart: Maybe worth mentioning that the hooks *can* be async, but don't have to be?

nbclient/client.py
davidbrochart: That would make sense, and we know if the cell is a Markdown cell with `cell.cell_type`.
But then maybe add `on_cell_execute` right before the execution? It won't be called if the cell is not a code cell or if the cell is skipped because of a tag.

devintang3: I moved `on_cell_start` towards the top so it'll execute for non-code cells and added  a new hook called `on_cell_execute`

docs/client.rst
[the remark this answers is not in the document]
devintang3: Thanks @choldgraf! Much appreciated.

I removed the line about functions can be async. If it can be async or sync, then I don't see much point in mentioning that

nbclient/client.py
devintang3: Added `on_notebook_error`, although I'm not sure if I did that part correctly. The only time I've personally seen the notebook fail is when there's a RuntimeError, so that's where I've placed the hook.

choldgraf: fair enough - and feel free to push back if you think that this is "too many hooks all at once" 😅 on the one hand I like symmetry and intentional design for extension points. On the other hand, I also believe in not building things until somebody actively wants it to be built :-)  so if you think this is too much complexity, another option could be to just add comments for where you think the extra hooks could go, and wait to build them until a user explicitly asks for it.

docs/client.rst
```diff
@@ -96,6 +96,36 @@ on both versions. Here the traitlet ``kernel_name`` helps simplify and
 maintain consistency: we can just run a notebook twice, specifying first
 "python2" and then "python3" as the kernel name.
 
+Hooks before and after cell execution
+-------------------------------------
```
davidbrochart, line 100:
```suggestion
Hooks before and after notebook or cell execution
-------------------------------------------------
```

docs/client.rst
```diff
@@ -96,6 +96,36 @@ on both versions. Here the traitlet ``kernel_name`` helps simplify and
...
+There are several configurable hooks that allow the user to execute code before and
+after a cell is executed. Each one is configured with a function that will be called in its
+respective place in the cell execution pipeline.
```
davidbrochart, line 103:
```suggestion
after a notebook or a cell is executed. Each one is configured with a function that will be called in its
respective place in the execution pipeline.
```

docs/client.rst
```diff
@@ -96,6 +96,36 @@ on both versions. Here the traitlet ``kernel_name`` helps simplify and
...
+Here is the available hooks:
+
+- ``on_notebook_start`` will run when the notebook client is initialized, before any execution has happened.
+- ``on_notebook_complete`` will run when the notebook client has finished executing, after kernel cleanup.
+- ``on_notebook_error`` will run when the notebook client has encountered an exception before kernel cleanup.
+
+**Cell-level hooks**: These hooks are called with two parameters:
+
+- ``cell=NotebookNode``: a reference to the current cell.
+- ``cell_index=int``: the index of the cell in the current notebook's list of cells
```
davidbrochart, line 119:
```suggestion
- ``cell_index=int``: the index of the cell in the current notebook's list of cells.
```

nbclient/client.py
```diff
@@ -261,6 +272,87 @@ class NotebookClient(LoggingConfigurable):
...
+        ),
+    ).tag(config=True)
+
+    on_notebook_error: t.Optional[t.Callable] = Callable(
+        default_value=None,
+        allow_none=True,
+        help=dedent(
+            """
+            Called when the notebook encountered an error.
+            Called with kwargs `notebook`
```
davidbrochart, line 304:
```suggestion
            Called with kwargs `notebook`.
```

nbclient/client.py
```diff
@@ -261,6 +272,87 @@ class NotebookClient(LoggingConfigurable):
...
+        ),
+    ).tag(config=True)
+
+    on_notebook_complete: t.Optional[t.Callable] = Callable(
+        default_value=None,
+        allow_none=True,
+        help=dedent(
+            """
+            Called after kernel is cleaned up.
+            Called with kwargs `notebook`
```
davidbrochart, line 293:
```suggestion
            Called with kwargs `notebook`.
```

nbclient/client.py
```diff
@@ -261,6 +272,87 @@ class NotebookClient(LoggingConfigurable):
...
     kernel_manager_class: KernelManager = Type(config=True, help='The kernel manager class to use.')
 
+    on_notebook_start: t.Optional[t.Callable] = Callable(
+        default_value=None,
+        allow_none=True,
+        help=dedent(
+            """
+            Called after the kernel manager and kernel client are setup, and cells
+            are about to execute.
+            Called with kwargs `notebook`
```
davidbrochart, line 282:
```suggestion
            Called with kwargs `notebook`.
```

nbclient/client.py
```diff
@@ -513,10 +607,13 @@ def on_signal():
             await self.async_start_new_kernel_client()
         try:
             yield
+        except RuntimeError:
+            await run_hook(self.on_notebook_error, notebook=self.nb)
+            raise
```
davidbrochart, line 612:
```suggestion
        except RuntimeError as e:
            await run_hook(self.on_notebook_error, notebook=self.nb)
            raise e
```

nbclient/tests/test_client.py
```diff
@@ -345,11 +345,7 @@ def test_async_parallel_notebooks(capfd, tmpdir):
     res = notebook_resources()
 
     with modified_env({"NBEXECUTE_TEST_PARALLEL_TMPDIR": str(tmpdir)}):
-        tasks = [
-            async_run_notebook(input_file.format(label=label), opts, res) for label in ("A", "B")
-        ]
-        loop = asyncio.get_event_loop()
-        loop.run_until_complete(asyncio.gather(*tasks))
+        [async_run_notebook(input_file.format(label=label), opts, res) for label in ("A", "B")]
```
davidbrochart, line 348: You removed the execution of these tasks, can you explain why?

nbclient/tests/test_client.py
```diff
@@ -370,9 +366,7 @@ def test_many_async_parallel_notebooks(capfd):
     # run once, to trigger creating the original context
     run_notebook(input_file, opts, res)
 
-    tasks = [async_run_notebook(input_file, opts, res) for i in range(4)]
-    loop = asyncio.get_event_loop()
-    loop.run_until_complete(asyncio.gather(*tasks))
+    [async_run_notebook(input_file, opts, res) for i in range(4)]
```
davidbrochart, line 369: Same here.

davidbrochart (changes requested):

nbclient/tests/test_client.py
devintang3: I think this was a mistake on my part when I was trying to investigate some failures. I've added those changes back in.

davidbrochart (approved):
Thanks a lot @devintang3, that looks good to me. A couple of notes, but I think this PR can be merged anyway:
- maybe use in the tests: `hooks  = [MagicMock() for i in range(7)]`.
- have all the hook help strings start with `A callable which executes...`.

devintang3:
> * maybe use in the tests: `hooks  = [MagicMock() for i in range(7)]`.
> * have all the hook help strings start with `A callable which executes...`.

That makes sense to me. I've updated the tests to do a loop and updated the help strings as well

davidbrochart (approved):

davidbrochart:
Thanks @devintang3 !

choldgraf:
🚀

PR #190, opened 2022-01-05
Compatibility with IPython 8 where tracebacks are different

### LINKED ISSUE #189: Compatibility with IPython 8 beta
Hello.

I've started testing IPython 8 beta release and I see that there is one incompatibility in tests, the problem is caused by this change: https://github.com/ipython/ipython/blob/master/docs/source/whatsnew/version8.rst#traceback-improvements which makes this regex https://github.com/jupyter/nbclient/blob/612a9e825a894498c8ca36a139fa73700db26910/nbclient/tests/test_client.py#L30 not compatible with the new output.

The whole output is kinda big but only three tests are failing and all of them for the same reason:
```
=========================== short test summary info ============================
FAILED nbclient/tests/test_client.py::test_run_all_notebooks[Interrupt.ipynb-opts6]
FAILED nbclient/tests/test_client.py::test_run_all_notebooks[Skip Exceptions with Cell Tags.ipynb-opts8]
FAILED nbclient/tests/test_client.py::test_run_all_notebooks[Skip Exceptions.ipynb-opts9]
```

Would it make sense to have a different regex for IPython 8+?

frenzymadness:
Fixed #189 

We cannot use different regex for different IPython versions because the testing files contain the old form so we have to use them both. All tests are passing now with IPython 8.


nbclient/tests/test_client.py
```diff
@@ -201,10 +204,10 @@ def normalize_output(output):
...
     if 'traceback' in output:
-        tb = [
-            re.sub(ipython_input_pat, '<IPY-INPUT>', strip_ansi(line))
-            for line in output['traceback']
-        ]
+        tb = []
+        for line in output["traceback"]:
+            line = re.sub(ipython_input_pat, '<IPY-INPUT>', strip_ansi(line))
+            line = re.sub(ipython8_input_pat, '<IPY-INPUT>', strip_ansi(line))
         output['traceback'] = tb
```
davidbrochart, line 211: `tb` is always empty here?

frenzymadness: I don't understand the question. The previous version created the content of `tb` as a list comprehension so the variable was and still is initiated here as a list of lines.

davidbrochart: But I only see `tb = []`, or am I missing something?

frenzymadness: Ah, I'm sorry, copy-paste error. Fixed. Thanks!

nbclient/tests/test_client.py
```diff
@@ -201,10 +204,11 @@ def normalize_output(output):
...
     if 'traceback' in output:
-        tb = [
-            re.sub(ipython_input_pat, '<IPY-INPUT>', strip_ansi(line))
-            for line in output['traceback']
-        ]
+        tb = []
+        for line in output["traceback"]:
+            line = re.sub(ipython_input_pat, '<IPY-INPUT>', strip_ansi(line))
+            line = re.sub(ipython8_input_pat, '<IPY-INPUT>', strip_ansi(line))
+            tb.append(line)
```
davidbrochart, line 211:
```suggestion
        tb = [
            re.sub(ipython8_input_pat, '<IPY-INPUT>', strip_ansi(
                re.sub(ipython_input_pat, '<IPY-INPUT>', strip_ansi(line))
            )
            for line in output['traceback']
        ]
```

frenzymadness: Well, honestly, my implementation seems to be more readable to me but it's your codebase so, if you want to, apply this suggestion. It might be also possible to merge the two regexes to one but again, readability would suffer.

davidbrochart: It was just a suggestion, I have no strong preference.

davidbrochart:
Thanks!

Issue #191, opened 2022-01-10
Compatibility of tests with ipywidgets 8 beta 1


frenzymadness:
Hello.

I'm testing also ipywidgets 8 beta 1 and it seems that there is only one problem. The new version of ipywidgets produces a different output in the notebook than what we expect in `test_run_all_notebooks[JupyterWidgets.ipynb-opts7]`:
```
>           assert normalized_expected_outputs == normalized_actual_outputs
E           AssertionError: assert [{'data': {'application/vnd.jupyter.widget-view+json': {'model_id': '<MODEL_ID>',\n                                                        'version_major': 2,\n                                                        'version_minor': 0},\n           'text/plain': 'LabelvalueHelloWorld'},\n  'output_type': 'display_data'}] == [{'data': {'application/vnd.jupyter.widget-view+json': {'model_id': '<MODEL_ID>',\n                                                        'version_major': 2,\n                                                        'version_minor': 0},\n           'text/plain': 'LabelvalueHelloWorld'},\n  'execution_count': 1,\n  'output_type': 'execute_result'}]
E             At index 0 diff: {'data': {'application/vnd.jupyter.widget-view+json': {'model_id': '<MODEL_ID>', 'version_major': 2, 'version_minor': 0}, 'text/plain': 'LabelvalueHelloWorld'}, 'output_type': 'display_data'} != {'output_type': 'execute_result', 'data': {'text/plain': 'LabelvalueHelloWorld', 'application/vnd.jupyter.widget-view+json': {'version_major': 2, 'version_minor': 0, 'model_id': '<MODEL_ID>'}}, 'execution_count': 1}
E             Full diff:
E               [
E                {'data': {'application/vnd.jupyter.widget-view+json': {'model_id': '<MODEL_ID>',
E                                                                       'version_major': 2,
E                                                                       'version_minor': 0},
E                          'text/plain': 'LabelvalueHelloWorld'},
E             +   'output_type': 'display_data'},
E             -   'execution_count': 1,
E             -   'output_type': 'execute_result'},
E               ]

nbclient/tests/test_client.py:231: AssertionError
```

martinRenou:
I believe this has to do with the change in https://github.com/jupyter-widgets/ipywidgets/pull/2021 and that it's an expected diff. So we'll need to change the test in nbclient (although I am not sure why there is a change in the execution count).

I guess this is only an issue with the test itself, and it shouldn't prevent nbclient to work with ipywidgets.

[...]

PR #192, opened 2022-01-10
Fix ipywidgets version in tests

martinRenou:
Fix https://github.com/jupyter/nbclient/issues/191

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,6 @@
 ipython
 ipykernel
-ipywidgets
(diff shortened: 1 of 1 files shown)


martinRenou:
This is temporary, the https://github.com/jupyter/nbclient/blob/master/nbclient/tests/files/JupyterWidgets.ipynb Notebook outputs should probably be regenerated when ipywidgets 8 is out

davidbrochart:
Thanks!

PR #195, opened 2022-01-14
Clear execution metadata, prefer msg header date when recording times

kevin-bates:
Hello,
While troubleshooting https://github.com/elyra-ai/elyra/issues/2387 and discussing the issue in https://github.com/deshaw/jupyterlab-execute-time/issues/60 it's probably a good idea to clear the cell metadata execution stanza prior to execution (which is the case in Lab).  This is due to the fact that nbclient doesn't post a value for `shell.execute_reply.started` which results in incorrect duration calculations by the execute-time extension when the notebook was first executed via Lab, then an nbclient application (like papermill).   Since the extension will fall back to using `iopub.execute_input` as its calculation for the cell's execution _start time_ in the absence of `shell.execute_reply.started`, clearing the stanza is sufficient to produce valid duration results.  I suspect this value is not captured because this is not specified in the NB format spec and, I'm guessing, this repo is trying to (correctly) adhere to the spec as much as possible.

The other observation is that Lab seeds this stanza's entries from the date in the corresponding message header rather than the current time, thereby removing network latency from the captured information.

This pull request clears the execution metadata when executing a cell.  It also _prefers_ the date-time from the corresponding message header, falling back to the current time if issues are encountered.  (I found that the mocking used in the tests results in "messages" that don't contain a header entry.  I went ahead and updated the timestamp function to tolerate that condition.)

If there's heartburn over the timestamp/header stuff, I'm okay retracting that, although it seems like the correct thing to do in light of the greater possibility of remotely-located kernels.

Looks like the traceback produced during the interrupt test has changed, causing some test failures.  Looking at existing PRs, this appears to be unrelated to this PR.

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -9,6 +9,7 @@ from contextlib import asynccontextmanager, contextmanager
 from queue import Empty
 from textwrap import dedent
 from time import monotonic
+from typing import Optional
 
 from jupyter_client import KernelManager
 from jupyter_client.client import KernelClient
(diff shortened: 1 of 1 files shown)


davidbrochart:
Thanks a lot Kevin, that looks good to me.

> Looks like the traceback produced during the interrupt test has changed, causing some test failures. Looking at existing PRs, this appears to be unrelated to this PR.

Yes, it is unrelated. I opened https://github.com/ipython/ipykernel/issues/845 but between IPython 8, ipykernel 6.7.0 and jupyter-client 7.1.1, I'm not sure.

PR #219, opened 2022-04-11
Add missing dep on testpath

blink1073:
cf https://github.com/jupyterlab/maintainer-tools/runs/5979418647?check_suite_focus=true


blink1073:
Typing errors are handled in #220

PR #218, opened 2022-04-05
Add more pre-commit hooks and update flake8

blink1073:

PR #220, opened 2022-04-12
Fix typings and update mypy settings

blink1073:
- Update for released typings in `jupyter_client`
- Use stricter mypy rules
- Include `py.typed` file for downstream typing support
- Retry last failed once on pytest error


blink1073:
@davidbrochart I plan to make a release with this and #218 tomorrow.

PR #224, opened 2022-04-28
Add error_on_interrupt trait

davidbrochart:
cc @brichet


brichet:
Thanks @davidbrochart, that works fine when timeout on kernel occurred.
As it is useful on timeout reply from kernel only (I think), shouldn't we rename the parameter to `error_on_timeout` instead of `error_on_interrupt` ?

brichet:
:+1:

PR #230, opened 2022-05-04
Clean up docs and typings

blink1073:
Fixes bug seen while working on https://github.com/jupyter/nbconvert/pull/1767

Also added `autodoc-traits`, but it currently doesn't handle converting help->docstrings.  I need to submit a patch there.

Also updates typings to handle types from `jupyter_client` and new rules in `mypy 0.950`.


blink1073:
Windows Python 3.10 failure is unrelated.

PR #234, opened 2022-05-31
Make sure kernel is cleaned up in case an error occurred while starting kernel client

CiprianAnton:
In case an error happens in `start_channels()`, the kernel is never cleaned up, leading to leaks. For e.g., in `GatewayKernelClient` there is a chance we get `ConnectionError`.

Another place to ensure cleanup is in `setup_kernel`, however, since `async_start_new_kernel_client` already performs cleanup in case of timeout, I think it's a better fit here.

Added test to cover this. All tests passed locally.

### DIFF
--- a/nbclient/cli.py
+++ b/nbclient/cli.py
@@ -35,12 +35,12 @@ class NbClientApp(JupyterApp):
     An application used to execute notebook files (``*.ipynb``)
     """
 
-    version = __version__
+    version = Unicode(__version__)
     name = 'jupyter-execute'
     aliases = nbclient_aliases
     flags = nbclient_flags
 
-    description = Unicode("An application used to execute notebook files (*.ipynb)")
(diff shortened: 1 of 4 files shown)


davidbrochart (approved):
Thanks a lot @CiprianAnton!
Let's just pin `traitlets<5.2.0` for now, until https://github.com/ipython/traitlets/issues/736 is fixed.
cc @blink1073

davidbrochart:
`traitlets` v5.2.2 was released.
I added some `type:ignore` for now because we need to have a better solution [in jupyter_client](https://github.com/jupyter/jupyter_client/pull/794) so that a `KernelClient` has a complete API. Currently only the subclasses `BlockingKernelClient` and `AsyncKernelClient` have a complete API, but it's hard to reconcile in `KernelClient` because methods can be async or not.

PR #237, opened 2022-06-28
Prevent file descriptor leak in util.just_run

mariokostelac:
While working with [papermill](https://papermill.readthedocs.io/en/latest/) I've found that repeated running of notebooks is eventually bumping into the limit of number of opened files.

Following he stacktrace led me to `util.just_run` function that creates asyncio loop in some cases, but doesn't close them, which leaves some `eventpoll` file descriptors opened forever.

This PR consists of two commits
- The first commit creates a failing test, showing that the current implementation is leaving file descriptors opened after each run. On my machine (unix selectors), it leaves 3 FDs per `just_run` call. It also explicitly declares `psutil` as a dependency. It was already a transitive dependency, but I'm promoting to be explicit since I'm calling it directly.
- The second commit fixes the issue by stopping and closing used loop, but just in cases where the method created the loop.

Happy to discuss the approach, missing context, and next steps to fix this bug.

### DIFF
--- a/nbclient/tests/test_util.py
+++ b/nbclient/tests/test_util.py
@@ -1,4 +1,5 @@
 import asyncio
+import unittest.mock
 from unittest.mock import MagicMock
 
 import psutil
@@ -94,3 +95,18 @@ def test_just_run_doesnt_leak_fds():
         just_run(async_sleep())
         diff.append(proc.num_fds() - fds_count)
     assert diff == [0] * 10
+
+
+def test_just_run_clears_new_loop():
+    async def async_sleep():
+        await asyncio.sleep(0.1)
+
+    loop = asyncio.new_event_loop()
+    loop.stop = MagicMock(wraps=loop.stop)
+    loop.close = MagicMock(wraps=loop.close)
+
+    with unittest.mock.patch.object(asyncio, "new_event_loop", return_value=loop):
+        just_run(async_sleep())
+
+    loop.stop.assert_called_once
+    loop.close.assert_called_once
--- a/nbclient/tests/test_util.py
+++ b/nbclient/tests/test_util.py
@@ -82,12 +82,9 @@ async def test_run_hook_async():
 def test_just_run_doesnt_leak_fds():
     proc = psutil.Process()
 
-    async def async_sleep():
-        await asyncio.sleep(0.01)
-
     # Warmup, just to make sure we're not failing on some initial fds being opened for the first time.
     for _ in range(10):
-        just_run(async_sleep())
+        just_run(asyncio.sleep(0.01))
     fds_count = proc.num_fds()
 
     diff = []
--- a/nbclient/tests/test_util.py
+++ b/nbclient/tests/test_util.py
@@ -1,10 +1,11 @@
 import asyncio
 from unittest.mock import MagicMock
 
+import psutil
 import pytest
 import tornado
 
-from nbclient.util import run_hook, run_sync
+from nbclient.util import just_run, run_hook, run_sync
 
 
 @run_sync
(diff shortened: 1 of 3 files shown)


mariokostelac:
I see some type checks are failing. I'll probably need help to get around it. I've tried using MagicMock, but some code asserts that loop objects is `AbstractEventLoop`, which makes it fail.

nbclient/tests/test_util.py
```diff
@@ -75,3 +77,36 @@ async def test_run_hook_async():
...
+        await asyncio.sleep(0.01)
+
+    # Warmup, just to make sure we're not failing on some initial fds being opened for the first time.
+    for _ in range(10):
+        just_run(async_sleep())
+    fds_count = proc.num_fds()
+
+    diff = []
+    for _ in range(10):
+        just_run(async_sleep())
```
davidbrochart, line 95: Maybe not worth defining a function?
```suggestion
        just_run(asyncio.sleep(0.01))
```

nbclient/tests/test_util.py
```diff
@@ -75,3 +77,36 @@ async def test_run_hook_async():
...
+    assert diff == [0] * 10
+
+
+def test_just_run_clears_new_loop():
+    async def async_sleep():
+        await asyncio.sleep(0.1)
+
+    loop = asyncio.new_event_loop()
+    loop.stop = MagicMock(wraps=loop.stop)
+    loop.close = MagicMock(wraps=loop.close)
```
davidbrochart, line 106: Let's just ignore type checks on these.
```suggestion
    loop.stop = MagicMock(wraps=loop.stop)  # type: ignore
    loop.close = MagicMock(wraps=loop.close)  # type: ignore
```

davidbrochart (changes requested):
Thanks a lot @mariokostelac, that looks good.

nbclient/tests/test_util.py
mariokostelac: I think I might have a better solution for these, without ignoring type checks.

I can't run tests locally, though.
```
(venv) dev-envs-ad\mario@a-2zurb4h4bbny ➜  nbclient git:(mario/fix_fd_leak) ls
binder  CHANGELOG.md  CONTRIBUTING.md  docs  LICENSE  MANIFEST.in  nbclient  nbclient.egg-info  pyproject.toml  pytest.ini  README.md  RELEASING.md  requirements-dev.txt  requirements.txt  setup.cfg  setup.py  venv
(venv) dev-envs-ad\mario@a-2zurb4h4bbny ➜  nbclient git:(mario/fix_fd_leak) tox -e py38
ERROR: tox config file (either pyproject.toml, tox.ini, setup.cfg) not found
```

davidbrochart: We're not using `tox`, you should just run `pytest`.

mariokostelac: Ah I see, I was running `pytest`, but it was missing the type checking piece that failed. I think https://github.com/jupyter/nbclient/blob/main/CONTRIBUTING.md might need an update.

Re: type checking. Would it make sense to put it as pre-commit check?

mariokostelac:
@davidbrochart I think this is in better shape now.

davidbrochart:
Trigger CI.

mariokostelac:
This is interesting. It seems that something in `jupyter_client` is now failing.
I've taken a look  at the stacktrace, it might be that  https://github.com/jupyter/jupyter_client/blob/3697a0d5e528be4c61fdfaaec0923acbfe0c1ea8/jupyter_client/utils.py#14 or https://github.com/jupyter/jupyter_client/blob/3697a0d5e528be4c61fdfaaec0923acbfe0c1ea8/jupyter_client/utils.py#18 is getting our closed loop, and trying to run on it. 

It's also strange that the function I'm changing looks remarkably similar to the one that's in the stacktrace. Why is that? @davidbrochart  I see that you've authored both of them (at least parts).

Also sorry for not running the whole testsuite and delaying this bug discovery till CI run.

I'll do some debugging tomorrow or on Friday. If you have some context that might help, I'm all ears.

mariokostelac:
I've tried hiding our loop from `jupyter_client` entirely by running `asyncio.set_event_loop(None)` after https://github.com/jupyter/nbclient/pull/237/files#diff-ce3eb5ea15b9f32af7e1dc032980287831900dc5538d389fc9875d7133c52220R63.

Unfortunately, https://github.com/jupyter/nbclient/blob/33b6e5fca02d6ede72643941fe0b8ec39d9c2437/nbclient/tests/test_client.py#L488 (and test after) then fail because `kc.shutdown` eventually tries to send something to control channel, which expect some running loop.
I'm not 100% sure I can locate the bug. Is the bug in the test, running kc.shutdown outside of the loop?

blink1073:
I like the idea of using `jupyter_client.utils.run_sync` here, and fixing it properly upstream.

Issue #244, opened 2022-08-10
Test failure in ipywidgets 8


jasongrout:
@frenzymadness [reported](https://github.com/jupyter-widgets/ipywidgets/issues/2750#issuecomment-1209792078) that nbclient tests fail when using the upcoming ipywidgets 8. The test notebook should be updated to reflect the change, as it is an expected change in ipywidgets behavior in ipywidgets 8.

See https://github.com/jupyter-widgets/ipywidgets/issues/2750#issuecomment-1209792078 and https://github.com/jupyter-widgets/ipywidgets/issues/2750#issuecomment-1210026954 for a short discussion of the issue.

jasongrout:
For convenience, here is the failure:

```
=================================== FAILURES ===================================
______________ test_run_all_notebooks[JupyterWidgets.ipynb-opts7] ______________

input_name = 'JupyterWidgets.ipynb', opts = {'kernel_name': 'python'}

    @pytest.mark.parametrize(
        ["input_name", "opts"],
        [
            ("Other Comms.ipynb", dict(kernel_name="python")),
            ("Clear Output.ipynb", dict(kernel_name="python")),
            ("Empty Cell.ipynb", dict(kernel_name="python")),
            ("Factorials.ipynb", dict(kernel_name="python")),
            ("HelloWorld.ipynb", dict(kernel_name="python")),
            ("Inline Image.ipynb", dict(kernel_name="python")),
            (
                "Interrupt.ipynb",
                dict(kernel_name="python", timeout=1, interrupt_on_timeout=True, allow_errors=True),
            ),
            ("JupyterWidgets.ipynb", dict(kernel_name="python")),
            ("Skip Exceptions with Cell Tags.ipynb", dict(kernel_name="python")),
            ("Skip Exceptions.ipynb", dict(kernel_name="python", allow_errors=True)),
            ("Skip Execution with Cell Tag.ipynb", dict(kernel_name="python")),
            ("SVG.ipynb", dict(kernel_name="python")),
            ("Unicode.ipynb", dict(kernel_name="python")),
            ("UnicodePy3.ipynb", dict(kernel_name="python")),
            ("update-display-id.ipynb", dict(kernel_name="python")),
            ("Check History in Memory.ipynb", dict(kernel_name="python")),
        ],
    )
    def test_run_all_notebooks(input_name, opts):
        """Runs a series of test notebooks and compares them to their actual output"""
        input_file = os.path.join(current_dir, 'files', input_name)
[...]

PR #246, opened 2022-08-22
Fix tests for ipywidgets 8

frenzymadness:
I believe that there is no reason to not use the latest versions of ipython and ipywidgets. Will see what CI thinks about it.


frenzymadness:
Fixes #244

frenzymadness:
First, CI failed completely, because there was missing dependency on nbconvert. Now, the failure of CI seems unrelated to this PR.

davidbrochart:
Yes, `test_many_parallel_notebooks` has always been flaky.
Thanks a lot @frenzymadness !

PR #250, opened 2022-09-07
Don't use nest-asyncio

davidbrochart:
Instead of using `nest-asyncio`, use something similar to @maartenbreddels's [with_event_loop](https://github.com/vaexio/vaex/blob/633970528cb5091ef376dbca2e4721cd42525419/packages/vaex-core/vaex/asyncio.py#L6-L19).


blink1073:
Replaced by #259

PR #251, opened 2022-09-09
Fix tests compatibility with IPython 8.5.0

frenzymadness:
The output has been changed again in IPython 8.5.0 https://github.com/ipython/ipython/blob/main/docs/source/whatsnew/version8.rst#restore-line-numbers-for-input


davidbrochart:
Thanks @frenzymadness, should we pin `ipython>=8.5.0` in [requirements-dev.txt](https://github.com/jupyter/nbclient/blob/main/requirements-dev.txt)?

frenzymadness:
> Thanks @frenzymadness, should we pin `ipython>=8.5.0` in [requirements-dev.txt](https://github.com/jupyter/nbclient/blob/main/requirements-dev.txt)?

That should not be needed. The regexes are compatible with older IPythons as well. If you'd like to simplify the code, we can do it in the future.

davidbrochart:
Great!

PR #253, opened 2022-09-21
Handle client 8 support

blink1073:
Testing against https://github.com/jupyter/jupyter_client/pull/835 (as well as client 7 to double-check).  Adds support for using synchronous clients since nbconvert will need it going forward.


blink1073:
@davidbrochart any object to making a release with this early next week?

blink1073:
I can make the release

PR #257, opened 2022-11-19
CI Refactor

blink1073:
Brings this repo in line with `jupyter_server`:

Update to use hatch and hatch envs
Remove use of circleCI
Clean up and improve GitHub Workflows
Add coverage handling
Improve mypy support

PR #259, opened 2022-11-24
Remove nest-asyncio

davidbrochart:
Experimenting with jupyter-client's [run_sync](https://github.com/jupyter/jupyter_client/blob/76a0f575d1e47b28f017a909df9c5b696a8207e8/jupyter_client/utils.py#L51) which doesn't use `nest-asyncio`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,7 +42,7 @@ repos:
     rev: v0.990
     hooks:
       - id: mypy
-        additional_dependencies: [jupyter_client, jupyter_core, nbformat, nbconvert]
+        additional_dependencies: [jupyter_client, jupyter_core>=5.1, nbformat, nbconvert]
 
   - repo: https://github.com/abravalheri/validate-pyproject
     rev: v0.10.1
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -5,6 +5,6 @@ dependencies:
   - numpy
   - pandas
   - matplotlib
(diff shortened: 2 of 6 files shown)


davidbrochart:
cc @maartenbreddels @SylvainCorlay

blink1073:
I'm hesitant to add a whole new library.  Perhaps we could put the code in Jupyter Core 5.x and 4.x?

davidbrochart:
Definitely not a new library just for this. Actually, it may also be fine like that, even though there's a bit of code duplication.

blink1073:
`jupyter_client` could depend on `jupyter_core` once it is there.

davidbrochart:
OK, I'll open PRs.

blink1073:
We should be able to use meeseeks to backport the PR from main.

davidbrochart:
This PR needs https://github.com/jupyter/jupyter_core/pull/315.

davidbrochart:
https://github.com/jupyter/nbclient/pull/259/commits/b77eb065e3db8bf7e1fa93d975b2d3e29cc68aab reapplies #258, which was overridden by #257.

blink1073:
We shouldn't per-emptively add upper bounds to dependencies, see https://iscinumpy.dev/post/bound-version-constraints/

davidbrochart:
> We shouldn't per-emptively add upper bounds to dependencies, see https://iscinumpy.dev/post/bound-version-constraints/

:+1:

PR #307, opened 2024-02-29
Optionally write out executed notebook in jupyter-execute

### LINKED ISSUE #306: Calling `jupyter-execute` runs the notebook but doesn't save it
I'm trying to execute notebooks from the command line using jupyter-execute with the minimum number of dependencies (I just want to run and save the notebook, and `nbconvert` is overkill for this).  If I use an environment with only "nbclient" and "ipykernel", running `jupyter execute notebook.ipynb` runs the notebook, but does not save the results anywhere.  There's no call back to `nbformat.write` anywhere in `cli.py`!.

Is this intended behavior?  If not, I think adding `nbformat.write(nb, input_path)` after the notebook is executed will fix it.

Thanks!

wpk-nist-gov:
Fixes #306.

Now actually write out the executed notebook to the same path as the input.

### DIFF
--- a/nbclient/cli.py
+++ b/nbclient/cli.py
@@ -22,6 +22,7 @@ nbclient_aliases: dict[str, str] = {
     "timeout": "NbClientApp.timeout",
     "startup_timeout": "NbClientApp.startup_timeout",
     "kernel_name": "NbClientApp.kernel_name",
+    "output": "NbClientApp.output_base",
 }
 
 nbclient_flags: dict[str, typing.Any] = {
@@ -32,6 +33,14 @@ nbclient_flags: dict[str, typing.Any] = {
             },
         },
         "Errors are ignored and execution is continued until the end of the notebook.",
+    ),
+    "inplace": (
+        {
+            "NbClientApp": {
+                "inplace": True,
+            },
+        },
+        "Overwrite input notebook with executed results.",
     ),
 }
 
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
LGTM, thanks!

blink1073:
The failure seems relevant, though I'm not sure what is triggering it about the line:

```python
nbclient/cli.py:200: in run_notebook
    with input_path.open() as f:
...
E       DeprecationWarning: pathlib.Path.__enter__() is deprecated and scheduled for removal in Python 3.13; Path objects as a context manager is a no-op
```

wpk-nist-gov:
I think the error is due to how I mocked things.  Working on it...

wpk-nist-gov:
No clue why CI is/was failing for macos 3.10.  I merged in changes from main. Hopefully that fixes it...

blink1073:
> No clue why CI is/was failing for macos 3.10.

The macos tests have been flaky for some time now.

blink1073:
Thanks again!

wpk-nist-gov:
> Thanks again!

My pleasure!

nbclient/cli.py
```diff
@@ -136,16 +177,27 @@ def run_notebook(self, notebook_path: str) -> None:
         # Log it
         self.log.info(f"Executing {notebook_path}")
 
-        name = notebook_path.replace(".ipynb", "")
+        input_path = Path(notebook_path).with_suffix(".ipynb")
```
Mazyod, line 180: @wpk-nist-gov Pardon my potential ignorance, but doesn't this change the input behavior?
The code was initially stripping the suffix then adding it, but this change adds the suffix indiscriminately. Seems if I pass a path to a notebook, the ipynb suffix will be duplicated.

wpk-nist-gov: `with_suffix` respects if the path already has the extension.  For example:

```pycon
>>> from pathlib import Path
# with suffix already there, don't replace
>>> Path("a/b/c.ipynb").with_suffix(".ipynb")
PosixPath('a/b/c.ipynb')
# without suffix, add it
>>> Path("a/b/c").with_suffix(".ipynb")
PosixPath('a/b/c.ipynb')
```

The previous behavior was to replace, then add the suffix:

```python
name = notebook_path.replace(".ipynb", "")

input_path = f"{name}.ipynb"
```

These should be mostly equivalent...

PR #310, opened 2024-04-15
Allow pytest 8

frenzymadness:
Flaky seems to be ready and the tests here seems to be compatible as well.


Carreau:
CLosing I think we are pytest >9 now anyway. Thanks.

PR #311, opened 2024-05-13
fix blocking issue when cell contains ipywidget

hjliu0206:
fix blocking issue when cell contains ipywidget


hjliu0206:
https://github.com/jupyter/nbclient/issues/312

hjliu0206:
@blink1073 
hi , blink, can you review my merge request ?

davidbrochart:
Thanks @hjliu0206 for opening an issue and a PR.
I'm not sure the fix should be here though, but rather upstream in [jupyter-client](https://github.com/jupyter/jupyter_client). It seems to me that `kernel_client.shell_channel.get_msg(timeout=None)` should return the message when it receives it, not after a timeout.

hjliu0206:
@davidbrochart hello , david, it's nbclient module block and pulls messages from the shell channel while sending messages to shell channel , most of the logic written in the nbclient , I'm not sure how to modify the jupyter client to fix this problem
image

hjliu0206:
If I make this change in my own company's branch of the project , does there exist any problems ?

davidbrochart:
> @davidbrochart If I make this change in my own company's branch of the project , do you think does there exist any potential problems ?

No, you are free to fork nbclient, but this might not be the best solution.
Maybe @maartenbreddels has ideas about the issue, since he's been involved in the output widget IIRC.

dleen:
Changing the code to `%matplotlib inline` (from `%matplotlib widget`) allows this example to succeed. So I suspect it is blocking due to some assumption in the `widget` engine

PR #327, opened 2025-04-16
Use filelock while starting a new kernel.

hodgestar:
I use nbmake and pytest-xdist to run many notebook tests in parallel on a CI machine with 64 cores. This frequency triggers the long standing `jupyter_client` bug https://github.com/jupyter/jupyter_client/issues/487 whose proposed solution is detailed in a Jupyter enhancement proposal accepted in 2023 -- https://github.com/jupyter/enhancement-proposals/pull/66.

When launch many local kernels, the obvious solution seems to be to hold a file lock during kernel start up. This PR implements that solution using https://github.com/tox-dev/filelock.

I've put up this PR without any tests yet as a request for feedback.

Questions:

- Should the filelock dependency be made optional? [Tentative answer: Probably]
- Should there be a filelock by default? [Tentative answer: Probably not -- the lock file path is likely not known at this level]
- What else would be needed to get this merged?

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -14,6 +14,7 @@ from queue import Empty
 from textwrap import dedent
 from time import monotonic
 
+import filelock
 from jupyter_client.client import KernelClient
 from jupyter_client.manager import KernelManager
 from nbformat import NotebookNode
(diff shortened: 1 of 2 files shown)


mschilli87:
I am hitting this using Snakemake's `notebook:` feature when running on several cores at once. If there is anything I can contribute to *finally* make Jupyter notebooks parallel-safe, please let me know.

Issue #329, opened 2025-07-14
Add on_cell_input Callable to Support Input


dgrahn:
Hello!

I am working on a project that requires executing Jupyter Notebooks from a Python script. `nbclient` would be ideal to do this, unfortunately the notebooks frequently request user input with `input` or `getpass`.  Per the [Jupyter Contribution Workflow](https://jupyter.readthedocs.io/en/latest/contributing/content-contributor.html), I am opening a proposal here.

I propose a new callable named `on_cell_input` is created. It is called with kwargs `cell`, `cell_index` and `cell_messages`. When this is set, the client will be updated to keep a log of cell messages and call the `on_cell_input` callable whenever an std message is requested. 

This is the code in question.

https://github.com/jupyter/nbclient/blob/2650b217c4c9965110ceb4de193092d7eb36c2bb/nbclient/client.py#L982-L986

It would be updated roughly as:

```py
if self.on_cell_input:
   run_cell = self.kc.execute(...)
   # handle messages
   parent_msg_id = await ensure_async(run_cell)
else:
   # Continue as usual
```

As this path will only be followed followed when `on_cell_input` is set, it should have minimal impact on anyone not using it (exactly one if-statement to check whether

davidbrochart:
Hi, this sounds like a nice feature, happy to review a PR!

PR #332, opened 2025-07-16
Add on_cell_input Callable to Support Input

dgrahn:
PR to support #329 

As few changes as possible were made.


dgrahn:
False alarm. I'll get this up and appropriate in a few minutes.

PR #334, opened 2025-07-24
Replace `@flaky.flaky` decorate with pytest marker

mgorny:
Use the `@pytest.mark.flaky` marker in place of the `@flaky.flaky` decorator, to modernize the code and improve compatibility with other plugins providing the feature such as `pytest-rerunfailures`.


Carreau:
I think it needs to be removed from `pyproject.toml` as well;

Carreau:
AH, no you still need the flake deps. All good then.

PR #338, opened 2025-12-22
Allow `display_id` to be `None`

davidbrochart:
See https://github.com/jupyter/nbclient/commit/3352484da606e1ab62705df7edd9ec1bc8f05262#r173257157.

### DIFF
--- a/nbclient/client.py
+++ b/nbclient/client.py
@@ -1152,7 +1152,6 @@ class NotebookClient(LoggingConfigurable):
             # if we have a hook registered, it will override our
             # default output behaviour (e.g. OutputWidget)
             hook = self.output_hook_stack[parent_msg_id][-1]
-            assert display_id is not None
             hook.output(outs, msg, display_id, cell_index)
             return None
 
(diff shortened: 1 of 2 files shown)


YannickJadoul:
In case it helped, I just bumped into this issue and created a minimal example:

Have a notebook with a single cell ([MWE.ipynb](https://github.com/user-attachments/files/24302109/MWE.ipynb))
```
import ipywidgets

def f(x):
    return x

w = ipywidgets.interact(f, x=ipywidgets.Dropdown(options=['a', 'b', 'c']))
```

Execute and convert like this:
```
$ jupyter nbconvert --execute --to markdown MWE.ipynb
```


I am not familiar with the inner workings of nbclient/nbconvert/ipywidgets/jupyter notebooks to know who's in the right here. So the error could potentially lie elsewhere than nbclient, for all I know (though curious to find out).
But it's a minimal example of what I bumped into, in case it's useful!

davidbrochart:
Thanks @YannickJadoul, your example works fine with this PR so I'll go ahead and merge it.

slayoo:
thank you!

PR #154, opened 2021-08-12
Use async generator backport only on old python

mkoeppe:
Fixes #153. Tests at https://github.com/mkoeppe/nbclient/actions/runs/1125181649

PR #340, opened 2026-03-14
drop leftover ref. to async_generator

a-detiste:
we are trying to remove async_generator from Debian.

this came up while grepping

this is a follow up to https://github.com/jupyter/nbclient/pull/154 from 2021


Carreau:
Thanks !

PR #345, opened 2026-06-03
no-numfocus funding + Zizmor

### LINKED ISSUE #344: Setup Zizmor and dependabot

Carreau:
Closes #344

### DIFF
--- a/.github/workflows/check-release.yml
+++ b/.github/workflows/check-release.yml
@@ -9,17 +9,25 @@ on:
   schedule:
     - cron: "0 0 * * *"
 
+permissions: {}
+
 jobs:
   check_release:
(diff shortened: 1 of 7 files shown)


.github/workflows/check-release.yml
krassowski, line 1: Should we just delete this thing? We already have another check release in `main.yaml`

Yann-P (approved):

.github/workflows/check-release.yml
Carreau: Sure, I opened and issue to do that later. I don't want to re-kick CI again.

krassowski:
Good to merge I think then!

PR #347, opened 2026-06-03
Minor cleanup and bump minimal versions.

Carreau:
I think jupyter core >=5.3.2 was a mistake and *strictly* was meant maybe?

And update all other deps to ~2 years old versions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,17 +29,12 @@ classifiers = [
     "License :: OSI Approved :: BSD License",
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.10",
-    "Programming Language :: Python :: 3.11",
-    "Programming Language :: Python :: 3.12",
-    "Programming Language :: Python :: 3.13",
-    "Programming Language :: Python :: 3.14",
 ]
 dependencies = [
-    "jupyter_client>=6.1.12",
-    "jupyter_core>=4.12,!=5.0.*",
-    "nbformat>=5.1.3",
-    "traitlets>=5.4",
+    "jupyter_client>=7.0.0",
+    "jupyter_core>=5.4.0",
+    "nbformat>=5.2.0",
+    "traitlets>=5.13",
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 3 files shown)


Carreau:
And after that I think we can release.

pyproject.toml
```diff
@@ -29,17 +29,12 @@ classifiers = [
...
 ]
 dependencies = [
-    "jupyter_client>=6.1.12",
-    "jupyter_core>=4.12,!=5.0.*",
-    "nbformat>=5.1.3",
-    "traitlets>=5.4",
+    "jupyter_client>=7.0.0",
+    "jupyter_core>=5.4.0",
+    "nbformat>=5.2.0",
+    "traitlets>=5.13",
```
krassowski, line 37: Why should we bump the minimum versions? We have a workflow job which ensures that `nblicent` tests pass with the minimum versions as currently declared.

krassowski (commented):
Otherwise I am ok to merge this (don't have rights to approve it).

pyproject.toml
Carreau: `>= 6.1.12` seem weird as its the last patch version of the 6.x branch. it seem weird to be compatbile with only a _single_ patch version for the major.  Either we are 6.x compatible with _some_ wiggle room on 6; or we just say that in practice we for all intent and purposes it's 7+. 

Same for nbformat.

And after it was pinning version for mypy and pre-commit (otherwise it's wild west and you don't know what you get), and if you do so, might as well sync the values with pyproject.toml, so have to pick some values. 

and while there, try to lipt a bit how far back we do traitlets. 

If we also release this to 0.11 (it's bee 0.10.x for a while), might as well bump now.

Carreau:
> Otherwise I am ok to merge this (don't have rights to approve it).

Gave you triage right, let me know if you want more.

krassowski (approved):
Let's try, we can relax the pins later if needed. In principle I would personally still keep the floor as low as tests require but do not see a strong reason against the proposal

krassowski:
@Carreau should we move forward here?

Carreau:
sounds good to me.

